PR #52, opened 2017-12-02
Ingest client

RussTorres:
Adding script based on Tim's current ingest worker which triggers an EM montage or Reference set by pointing at its directory.

### DIFF
--- a/rendermodules/ingest/ingest_tileset.py
+++ b/rendermodules/ingest/ingest_tileset.py
@@ -54,7 +54,7 @@ class IngestTileSetModule(argschema.ArgSchemaParser):
     lc_re = '_metadata_.*reference.*.json'
 
     # regex to find manifest file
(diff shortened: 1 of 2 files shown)


fcollman:
This looks ok to me... my only question is whether it belongs in this repo or not... this is the first module that pulls in blue sky as a dependency in this project.  So on a technical note the workflow_client should be added the the requirements.. and workflow_client is presently available externally.  I'm happy to have it here for now, but we should think about it.

RussTorres:
You're right -- this and any other ingest (based on microscope type) might be better to have in something like the at_em_blue-sky repo.  That said with render-modules' current setup it shouldn't be too far of a stretch to include workflow_client as an external dependency.

RussTorres:
Closing for now while waiting for workflow_client modifications which should allow mock.

PR #53, opened 2017-12-09
Mipmaps decmods

RussTorres:
Adds output schemas for mipmaps and uses tilespec-based parallelization for applymipmaps.  These versions were cherry picked from repo used for EM December SMART goal.

### DIFF
--- a/rendermodules/dataimport/create_mipmaps.py
+++ b/rendermodules/dataimport/create_mipmaps.py
@@ -1,10 +1,12 @@
-from PIL import Image
-import argparse
 import os
+from PIL import Image
 from rendermodules.module.render_module import RenderModuleException
+
 
 class CreateMipMapException(RenderModuleException):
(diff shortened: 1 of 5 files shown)


rendermodules/dataimport/create_mipmaps.py
```diff
@@ -1,10 +1,12 @@
-from PIL import Image
-import argparse
 import os
+from PIL import Image
 from rendermodules.module.render_module import RenderModuleException
 
+
 class CreateMipMapException(RenderModuleException):
     """Exception raised when there is a problem creating a mipmap"""
+    pass
```
fcollman: I'm not concerned about this, just curious... is this pass actually necessary?

RussTorres: I guess either that or a docstring does the trick (and docstring actually adds something), so there's not a need for both.  More force of habit than anything else.

PR #54, opened 2017-12-19
Em montage

mahaling:
modified to adapt to the latest version of the solver


fcollman:
I think we should close this in favor of the em_montage_integration_test branch

PR #56, opened 2017-12-21
Montage stack

fcollman:
This solves the problem of the injesting stack being in the loading stack by cloning the section to be montage solved into a temporary section only stack before solving and then deleting it after.  It is transparent to the workflow engine.  We can if we want explore in the future having it push the single section stack to another render engine to make there be no more net load on the render server (one read, one write, per section per solve).


fcollman:
closing in favor of #57

PR #57, opened 2017-12-21
Montage stack

fcollman:
This solves the problem of the injesting stack being in the loading stack by cloning the section to be montage solved into a temporary section only stack before solving and then deleting it after. It is transparent to the workflow engine. We can if we want explore in the future having it push the single section stack to another render engine to make there be no more net load on the render server (one read, one write, per section per solve).

### DIFF
--- a/rendermodules/montage/run_montage_job_for_section.py
+++ b/rendermodules/montage/run_montage_job_for_section.py
@@ -104,8 +104,9 @@ class SolveMontageSectionModule(RenderModule):
         self.args.pop('solver_executable', None)
         self.clone_section_stack = self.args['clone_section_stack']
         self.args.pop('clone_section_stack')
-        if (self.args['first_section']!=self.args['last_section']):
-            raise ValidationError("first section and last section not equal")
+        if self.clone_section_stack:
+            if (self.args['first_section']!=self.args['last_section']):
+                raise ValidationError("first section and last section not equal")
 
     def run(self):
 
@@ -146,7 +147,7 @@ class SolveMontageSectionModule(RenderModule):
         if self.args.get('output_json',None) is not None:
             sectionDataList = renderapi.stack.get_stack_sectionData(self.args['target_collection']['stack'],
                 render=self.render)
-            sectionData = next(section for section in sectionDataList if section['z']==self.args['first_section'])
+            sectionData = [section for section in sectionDataList if (section['z']>=self.args['first_section']) and (section['z'<=self.args['last_section'])]
             self.output(sectionData)
         
         #if you made a tmp stack destroy it
--- a/rendermodules/montage/run_montage_job_for_section.py
+++ b/rendermodules/montage/run_montage_job_for_section.py
@@ -153,7 +153,7 @@ class SolveMontageSectionModule(RenderModule):
             d={'zs':zs}
             self.output(d)
         except:
-            pass
+            raise RenderModuleException("unable to output json {}",d)
         
         #if you made a tmp stack destroy it
         if self.clone_section_stack:
--- a/rendermodules/montage/run_montage_job_for_section.py
+++ b/rendermodules/montage/run_montage_job_for_section.py
@@ -147,7 +147,9 @@ class SolveMontageSectionModule(RenderModule):
         if self.args.get('output_json',None) is not None:
             sectionDataList = renderapi.stack.get_stack_sectionData(self.args['target_collection']['stack'],
                 render=self.render)
-            sectionData = [section for section in sectionDataList if (section['z']>=self.args['first_section']) and (section['z'<=self.args['last_section'])]
+            sectionData = [section for section in sectionDataList \
+                if (section['z']>=self.args['first_section']) \
+                and (section['z']<=self.args['last_section'])]
             self.output(sectionData)
         
         #if you made a tmp stack destroy it
--- a/rendermodules/montage/run_montage_job_for_section.py
+++ b/rendermodules/montage/run_montage_job_for_section.py
@@ -118,7 +118,7 @@ class SolveMontageSectionModule(RenderModule):
                                            time.strftime("%m%d%y_%H%M%S"))
             zvalues = renderapi.stack.get_z_values_for_stack(self.args['source_collection']['stack'],render=self.render)
             zs = [z for z in zvalues if \
-                if (z>=self.args['first_section']) \
+                (z>=self.args['first_section']) \
                 and (z<=self.args['last_section'])]]
             renderapi.stack.clone_stack(self.args['source_collection']['stack'],
                                         tmp_stack,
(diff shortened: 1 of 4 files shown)


rendermodules/montage/run_montage_job_for_section.py
```diff
@@ -167,8 +102,22 @@ def __init__(self, schema_type=None, *args, **kwargs):
         # Assigning the solver_executable to a different variable and removing it from args
         self.solver_executable = self.args['solver_executable']
         self.args.pop('solver_executable', None)
+        self.clone_section_stack = self.args['clone_section_stack']
+        self.args.pop('clone_section_stack')
+        if (self.args['first_section']!=self.args['last_section']):
```
mahaling: this condition should be applied only when the clone_section_stack is set to true...

fcollman: i don't think that's right... this part needs to pop the parameter from the arguments, it has a value no matter what even if it's not set because i gave it a default in the schema.

mahaling: Yes, I agree with what you said... But I meant the "if" condition that checks for the self.args['first_section']!= self.args['last_section']. 
This condition is checked regardless of the value of "clone_section_stack". However, it should not be unless you are running this code for only one section. This might not be the case if you want to montage a range of sections  (for the subsequent time with different parameters using the source stack with "COMPLETE" state).

encapsulating this "if" condition (line 107) with another one that checks if "clone_section_stack" is set to true would be appropriate.

fcollman: ah! yes i see what you are saying now. You are right! i'll change this.

rendermodules/montage/run_montage_job_for_section.py
```diff
@@ -167,8 +102,23 @@ def __init__(self, schema_type=None, *args, **kwargs):
         # Assigning the solver_executable to a different variable and removing it from args
         self.solver_executable = self.args['solver_executable']
         self.args.pop('solver_executable', None)
+        self.clone_section_stack = self.args['clone_section_stack']
+        self.args.pop('clone_section_stack')
+        if self.clone_section_stack:
+            if (self.args['first_section']!=self.args['last_section']):
+                raise ValidationError("first section and last section not equal")
```
RussTorres: Is there a reason why we shouldn't let this work for multiple zs?  It could be helpful in manual/batch processing and shouldn't be a very different implementation.

rendermodules/montage/run_montage_job_for_section.py
```diff
@@ -167,8 +102,23 @@ def __init__(self, schema_type=None, *args, **kwargs):
...
 
     def run(self):
+
+        if self.clone_section_stack:
+            tmp_stack = "{}_z{}".format(self.args['source_collection']['stack'],self.args['first_section'])
```
RussTorres: Maybe the temporary stack name should have a randomized or timestamped component -- in the case of reimaging and possible metadata errors z index is not reliable and it's possible that some of these stacks are not deleted after creation (e.g. in case of errors raised somewhere else in the script).

rendermodules/montage/run_montage_job_for_section.py
```diff
@@ -180,32 +130,10 @@ def run(self):
...
-        # Note that MCRROOT is the matlab compiler runtime's root folder
-        '''
-            LD_LIBRARY_PATH=.:${MCRROOT}/runtime/glnxa64 ;
-            LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:${MCRROOT}/bin/glnxa64 ;
-            LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:${MCRROOT}/sys/os/glnxa64;
-            LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:${MCRROOT}/sys/opengl/lib/glnxa64;
-        '''
-
         if "MCRROOT" not in os.environ:
             raise ValidationError("MCRROOT not set")
```
RussTorres: Minor, but this error could be raised before generating a temporary file.

rendermodules/montage/run_montage_job_for_section.py
```diff
@@ -215,36 +143,19 @@ def run(self):
...
-                f.write('#PBS -q emconnectome\n')
-                f.write('%s\n'%(cmd_to_qsub))
-            f.close()
-
-            qsub_cmd = 'qsub %s'%(temppbs.name)
-            subprocess.call(qsub_cmd)
-        '''
+        #return the section data as the output_json if it was given
+        if self.args.get('output_json',None) is not None:
+            sectionDataList = renderapi.stack.get_stack_sectionData(self.args['target_collection']['stack'],
```
RussTorres: Is the solver setting the target collection to complete?  That was configurable, but I don't know if it made it into the dev branch.  I think this will break on LOADING stacks prohibiting continuous processing.

fcollman: I'm not sure.. but I think that's a change we should probably implement in the Solver, as we would have problems if the solver closed the stack and then this module reset it to loading.. that might leave some spare moments of closed stack that could cause problems, so i'm not going to make any changes here.

rendermodules/montage/run_montage_job_for_section.py
fcollman: i can't see a reason... i've changed the module to allow it

rendermodules/montage/run_montage_job_for_section.py
fcollman: done

rendermodules/montage/run_montage_job_for_section.py
mahaling: It was an option in the old version of the solver, but not the new one. But that can be changed easily. I will make that change internal to the solver... Which then requires this module to take that option as input as well. Or the other way is to leave the stack in LOADING state by default from the solver side.

fcollman: I think we should make the solver take this as an option, and then we make the default value in the schema here to be to leave it Loading.

fcollman: can we go ahead and approve this PR though? this change is needed for beta1 and having the stack in a complete state won't break future stacks from solving into it, as the solver sets it to loading before uploading results.

RussTorres: In the event of concurrent montaging having individual montage solves setting the stack to COMPLETE should prevent others which are in the process of uploading tiles, right?

fcollman:
Ah you are correct.

rendermodules/montage/run_montage_job_for_section.py
fcollman: yes you are right.. i've made the change to the modules and added the option in the solver.

fcollman:
This is ready now I think... the code coverage is less than target which is why there is no check mark..... but I don't see why... overall test coverage is improved.

PR #58, opened 2018-01-04
Spark logging

fcollman:
These give optional additional hooks to configure spark through the python interface, we are doing this to give the ability to dynamically configure logging, and the test is setup to demonstrate how to do that.

### DIFF
--- a/rendermodules/pointmatch/generate_point_matches_spark.py
+++ b/rendermodules/pointmatch/generate_point_matches_spark.py
@@ -43,7 +43,7 @@ example = {
 def add_arg(l,argname,args):
     value = args.get(argname,None)
     if value is not None:
-        l+=["--{}".format(argname),args[argname]]
+        l+=["--{}".format(argname),"{}".format(args[argname])]
 
 def form_sift_params_list(args):
     sift_params = []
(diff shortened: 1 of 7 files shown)


rendermodules/pointmatch/generate_point_matches_spark.py
```diff
@@ -78,11 +78,18 @@ def run(self):
         cmd = "{} --master {}".format(sparksubmit, self.args['masterUrl'])
         cmd = cmd + " --executor-memory {}".format(self.args['memory'])
         cmd = cmd + " --driver-memory {}".format(self.args['driverMemory'])
+        if self.args.get('spark_files',None) is not None:
+            for spark_file in self.args['spark_files']:
+                cmd = cmd + " --files {}".format(spark_file)
+        if self.args.get('spark_conf',None) is not None:       
+            for key,value in  self.args['spark_conf'].items():
+                cmd = cmd + " --conf {}='{}'".format(key,value)
```
RussTorres: The spark submit documentation suggests wrapping the entire `<key>=<value>` in double quotes in case of spaces, but a more readable alternative in python might be to use the subprocess module.

e.g. 
```subprocess.call(['bash', 'some file.sh'])```
will work while
```os.system('bash some file.sh')``` will fail looking for a file named "some"

fcollman:
OK.. i have added some mocked testing for the qsub module, it revealed a problem with the schema which i solved.  It doesn't actually check that the qsub command works.

PR #59, opened 2018-01-05
Mipmap coverage clean

fcollman:
in an effort to improve our code coverage i discovered some inconsistencies in the modules, mostly around exception handling and validation that wasn't being tested.  This PR cleans up some of those problems and improves the testing coverage.

### DIFF
--- a/rendermodules/dataimport/generate_mipmaps.py
+++ b/rendermodules/dataimport/generate_mipmaps.py
@@ -104,11 +104,10 @@ class GenerateMipMaps(RenderModule):
                     zvalues = [self.args['z']]
             except KeyError:
                 raise RenderModuleException('No z value given for mipmap generation')
-        self.logger.error("{}".format(zvalues))
-        if len(zvalues) == 0:
+        if not zvalues:
             raise RenderModuleException('No sections found for stack {}'.format(
                 self.args['input_stack']))
-
+        
         self.logger.debug("Creating mipmaps...")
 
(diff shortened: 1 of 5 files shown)


rendermodules/dataimport/generate_mipmaps.py
```diff
@@ -98,17 +98,16 @@ def run(self):
             zvalues1 = range(self.args['zstart'], self.args['zend'] + 1)
             # extract only those z's that exist in the input stack
             zvalues = list(set(zvalues1).intersection(set(zvalues)))
-        except NameError:
+        except KeyError:
             try:
                 if self.args['z'] in zvalues:
                     zvalues = [self.args['z']]
```
RussTorres: If you pick a z outside of this zvalues in the stack, does this generate mipmaps for the entire stack?

mahaling: The zvalues list should be empty in that case and hence the code should exit gracefully without generating any mipmaps

fcollman: Russel is right.. if you only put self.args['z'] and that z value isn't in the stack, then the if statement isn't called, and zvalues remains all the zvalues which isn't the right behavior.  I'll fix it with an else statement and add a test.

mahaling: ok got it.... I was referring to line 100...

PR #68, opened 2018-01-28
Rough align module

mahaling:
submitting this PR to check the integration test and its code coverage.


fcollman:
because this branch was inactive for over a month the bamboo build plan deactivated it.. i've manually reactivated it.

rendermodules/rough_align/apply_rough_alignment_to_montages.py
```diff
@@ -0,0 +1,244 @@
...
+            d1 = t.to_dict()
+            d1['z'] = newz
+            allts.append(d1)
+
+        tilespecfilename = os.path.join(output_dir,'tilespec_%04d.json'%newz)
+        fp = open(tilespecfilename,'w')
+        json.dump([ts for ts in allts], fp, indent=4)
+        fp.close()
+    except:
+        print("This z has not been aligned!")
```
fcollman: do we want this to proceed as if nothing happened if there is a z section that we ask to apply alignment to that has not been aligned?

integration_tests/test_rough_align.py
```diff
@@ -0,0 +1,382 @@
...
+                    renderapi.stack.get_stack_tileIds(
+                        test_montage_stack, render=render)))) == 0
+    
+    zvalues = render.run(renderapi.stack.get_z_values_for_stack,
+                          test_montage_stack)
+    zs = [1020, 1021, 1022]
+    assert(set(zvalues) == set(zs))
+
+    yield test_montage_stack
+    #renderapi.stack.delete_stack(test_montage_stack, render=render)
```
fcollman: i'm conflicted as to whether this should be uncommented.. on one hand its a nice idea to clean up after ourselves with tests and makes test easy to re-run without confusing results with the render server, on the other hand its annoying to debug what went wrong with the test when the stack is gone afterwards.  Interested to hear other people's thoughts about this... more than anything I think we should be consistent.

integration_tests/test_rough_align.py
```diff
@@ -0,0 +1,382 @@
...
+    renderapi.stack.set_stack_state(one_tile_montage_stack, 
+                                    'COMPLETE', 
+                                    render=render)
+    # assure stack is built correctly
+    assert len({tspec.tileId for tspec
+                in tspecs}.symmetric_difference(set(
+                    renderapi.stack.get_stack_tileIds(
+                        one_tile_montage_stack, render=render)))) == 0
+    yield one_tile_montage_stack
+
```
fcollman: delete stack issue is here too.

integration_tests/test_rough_align.py
```diff
@@ -0,0 +1,382 @@
...
+    }
+    ex['output_json'] = os.path.join(image_directory, 'output.json')
+
+    mod = RenderSectionAtScale(input_data=ex, args=[])
+    mod.run()
+
+    out_dir = os.path.join(image_directory, render_params['project'], montage_stack, 'sections_at_0.1/001/0')
+    assert(os.path.exists(out_dir))
+    
+    files = glob.glob(os.path.join(out_dir, '*.png'))
```
fcollman: should assert that it made the right number of files for the zrange

integration_tests/test_rough_align.py
```diff
@@ -0,0 +1,382 @@
...
+    pt_match_collection = 'rough_point_match_collection'
+    renderapi.pointmatch.import_matches(pt_match_collection,
+                                        rough_point_matches_from_json,
+                                        render=render)
+    
+    # check if point matches have been imported properly
+    groupIds = render.run(renderapi.pointmatch.get_match_groupIds, pt_match_collection)
+    #assert(len(groupIds) == 3)
+    yield pt_match_collection
+
```
fcollman: another place to consider the cleaning up issue.

integration_tests/test_rough_align.py
```diff
@@ -0,0 +1,382 @@
...
+    mod.run()
+
+    with open(ex['output_json'], 'r') as f:
+        js = json.load(f)
+    
+    out_dir = os.path.join(image_directory, render_params['project'], js['temp_stack'], 'sections_at_0.1/001/0')
+    assert(os.path.exists(out_dir))
+    
+    files = glob.glob(os.path.join(out_dir, '*.png'))
+    for fil in files:
```
fcollman: shouldn't this create one specific file?

integration_tests/test_rough_align.py
mahaling: This should not be commented... I am having it commented until I reach the code coverage.

integration_tests/test_rough_align.py
mahaling: same comment as above

integration_tests/test_rough_align.py
mahaling: will add it.

integration_tests/test_rough_align.py
mahaling: No, each section has its own downsample image if that's what you are referring to. Also, individual tilespec file.

rendermodules/rough_align/apply_rough_alignment_to_montages.py
mahaling: A potential case where you would want to simply pass is in case of a missing section... But if we are handling that in a different way then this will become an exception.

integration_tests/test_rough_align.py
mahaling: In this specific case... this comment is good as the stack is required till the end of the test and a reference once forwards the execution to the delete statement - not the right way I want the testing to be done.

rendermodules/rough_align/apply_rough_alignment_to_montages.py
fcollman: the missing section case is a good use case to consider.. doesn't the module code detect though what z values are present at the start of the module?  Another suggestion would be to have the partial mapped function return True/False per section, and then have the output json report the z values of the sections that were successfully solved.  In that way the strategy could make a call as to whether it should proceed or not.

mahaling: Technically rough alignment shouldn't be affected if there is a missing section. Hence the reason why I left the exception to be a print statement... But returning true/false seems good, Although I feel that there might be more use cases like no point matches found, etc...

RussTorres: agreed that bare try/except is scary.  Which exceptions are raised in this block -- RenderErrors? IOErrors?  It would be nice if we could differentiate between, say, a missing section and a failure to get stack bounds because either input_stack or prealigned_stack are in LOADING state.

rendermodules/rough_align/apply_rough_alignment_to_montages.py
```diff
@@ -0,0 +1,244 @@
+
+import os
+import json
+import renderapi
+import glob
+import numpy as np
+from ..module.render_module import RenderModule
+from rendermodules.rough_align.schemas import ApplyRoughAlignmentTransformParameters
+from rendermodules.stack.consolidate_transforms import *
```
RussTorres: for the sake of readability, it could be helpful be explicit here -- it's not obvious, for instance, that RenderModuleException is being imported from rendermodules.stack.consolidate_transforms

rendermodules/materialize/schemas.py
```diff
@@ -1,11 +1,60 @@
...
+        required=False,
+        default=20,
+        missing=20,
+        description='number of parallel threads to use')
+
+class RenderSectionAtScaleOutput(argschema.schemas.DefaultSchema):
+    image_directory = InputDir(
+        required=True,
+        description='Directory in which the downsampled section images are saved')
+    temp_stack = Str(
```
RussTorres: This is required, but doesn't seem to be used in the module -- the temporary stack name is derived from the input stack.

rendermodules/materialize/render_downsample_sections.py
```diff
@@ -0,0 +1,159 @@
...
+                        self.args['temp_stack'])
+
+            # set stack state to LOADING
+            self.render.run(renderapi.stack.set_stack_state,
+                        self.args['temp_stack'],
+                        state='LOADING')
+
+            # import json files into temp_no_mipmap_stack
+            # this also sets the stack's state to COMPLETE
+            self.render.run(renderapi.client.import_jsonfiles_parallel,
```
RussTorres: this could be slow for a small number of sections -- handling a long list of tilespecs rather than section-parallel jsons makes better use of the processing pool in the workflow's current use case.

rendermodules/materialize/render_downsample_sections.py
```diff
@@ -0,0 +1,159 @@
...
+                        self.args['temp_stack'],
+                        self.args['image_directory'],
+                        zvalues,
+                        scale=self.args['scale'],
+                        format=self.args['imgformat'],
+                        doFilter=self.args['doFilter'],
+                        fillWithNoise=self.args['fillWithNoise'])
+        
+        # delete the temp stack
+        #self.render.run(renderapi.stack.delete_stack, self.args['temp_stack'])
```
RussTorres: This should be uncommented for production.

rendermodules/materialize/schemas.py
```diff
@@ -1,11 +1,60 @@
...
+        missing=-1,
+        description='max Z to create the downsample section from')
+    pool_size = Int(
+        required=False,
+        default=20,
+        missing=20,
+        description='number of parallel threads to use')
+
+class RenderSectionAtScaleOutput(argschema.schemas.DefaultSchema):
+    image_directory = InputDir(
```
RussTorres: Should this be an OutputDir?

RussTorres: Whoops, misread this -- I see now that this is an output schema

rendermodules/materialize/render_downsample_sections.py
mahaling: I did not quite get this one. Are you saying that each section should have its own json?. If so, it is creating one json for each z. If its not the case, can you please elaborate.

rendermodules/rough_align/apply_rough_alignment_to_montages.py
mahaling: I guess its appropriate to do these checks before this function gets executed... There is definitely a possibility that the montage stack is in LOADING state in terms of how the workflow is designed....

rendermodules/materialize/render_downsample_sections.py
RussTorres: I think the way that import_jsonfiles_parallel works is to spawn a pool in which a single worker recieves a single jsonfile from your input list -- so in the example of running this for a single section you are spinning up a pool of n processes, but only using one to actually send anything to the webservice.  import_tilespecs_parallel takes as input a list of tilespecs and a poolsize and runs however many processes are in your pool as ingest clients running evenly divided subsets of the input tilespecs.

RussTorres: this is more of an optimization and definitely not blocking for this PR, it just means you don't have to track temporary jsons and speeds up processing.

rendermodules/rough_align/apply_rough_alignment_to_montages.py
RussTorres: Shouldn't the only expected failure be in `renderapi.stack.get_tile_specs_from_z`? If so, could failures in those just return None so anything else weird will just fail?
Also, it looks like `highres_ts` is only being used as a proxy for `renderapi.transform.AffineModel` and fetching all tilespecs in z can be expensive with large collections.

fcollman: I am still concerned about this bare try/except beyond expected failures, what if somehow the render server goes down or fails to communicate due to no problem of this code at all.. or nfs goes down, or who knows what.  I don't think we want the workflow plowing forward, it will just make it harder to figure out what happened and to fix it up later.

integration_tests/test_rough_align.py
fcollman: yield will cause the stack to stick around until all the tests that use this fixture are completed, so if you have a test that needs that stack to exist, simply referencing this feature will force it to stick around.  I wasn't sure if 'reference once forwards' means that you thought that the first time it was referenced it would proceed to delete?

PR #69, opened 2018-01-30
Materialization

RussTorres:
Materialize sections using betterboxclient.


RussTorres:
the materialize schemas might overlap with a file created in #68 ....

integration_tests/test_data.py
```diff
@@ -1,163 +1,173 @@
-import os
```
fcollman: I don't understand what happened here... the diff is nuts, I can't see what is new and what isn't..

rendermodules/materialize/schemas.py
```diff
@@ -0,0 +1,84 @@
...
+class MaterializedBoxParameters(argschema.schemas.DefaultSchema):
+    stack = Str(required=True, description=(""))
+    rootDirectory = OutputDir(required=True, description=(""))
+    width = Int(required=True, description="")
+    height = Int(required=True, description="")
+    maxLevel = Int(required=False, default=0, description="")
+    fmt = Str(required=False)  # TODO validate to allow PNG, TIF(F?), JPG
+    maxOverviewWidthAndHeight = Int(required=False, description=(""))
+    skipInterpolation = Boolean(required=False, description=(""))
+    binaryMask = Boolean(required=False, description=(""))
```
fcollman: I think this description as a tuple will break the documentation

integration_tests/test_data.py
RussTorres: ugh, yes.  I had to reinstall git and might have introduced a line ending mismatch or something.  I can close and roll back commits to try and triage.

rendermodules/materialize/schemas.py
RussTorres: not tuples, just empty strings.  Either way, I will need to commit to add documentation.

RussTorres:
Misfire.  Closing until the commit log and documentation is presentable.

PR #70, opened 2018-01-30
Materialization

RussTorres:
Once more, materialization based on BetterBoxClient


rendermodules/materialize/materialize_sections.py
```diff
@@ -0,0 +1,122 @@
...
+
+class MaterializeSectionsError(RenderModuleException):
+    pass
+
+
+class MaterializeSectionsModule(argschema.ArgSchemaParser):
+    default_schema = MaterializeSectionsParameters
+    default_output_schema = MaterializeSectionsOutput
+
+    @staticmethod
```
fcollman: this seems generic..should it be in a more generic location?

rendermodules/materialize/schemas.py
```diff
@@ -0,0 +1,121 @@
...
+        if data.get('baseDataUrl') is None:
+            data['baseDataUrl'] = '{host}{port}/render-ws/v1'.format(
+                host=(data['render']['host']
+                      if data['render']['host'].startswith(
+                      'http') else 'http://{}'.format(data['render']['host'])),
+                port=('' if data['render']['port'] is None else ':{}'.format(
+                      data['render']['port'])))
+
+
+class SparkParameters(argschema.schemas.DefaultSchema):
```
fcollman: This is also duplicated I think

RussTorres: This is the more general class of what is also implemented in the pointmatching.  Pointmatching currently implements this as part of a single schema class with module-specific parameters.  In a better world, parameters for spark modules would subclass something like this -- similar to what I have here.

rendermodules/materialize/materialize_sections.py
RussTorres: these more generic classes are based off the java client parameters and could be included in some other module which is imported --  maybe rendermodules.module?

rendermodules/materialize/schemas.py
```diff
@@ -0,0 +1,121 @@
...
+
+class RenderWebServiceParameters(argschema.schemas.DefaultSchema):
+    baseDataUrl = Str(required=False, description=(
+        "api endpoint url e.g. http://<host>[:port]/render-ws/v1"))
+    owner = Str(required=False, description=("owner of target collection"))
+    project = Str(required=False, description=("project fo target collection"))
+
+
+class RenderParametersRenderWebServiceParameters(RenderWebServiceParameters):
+    render = Nested(
```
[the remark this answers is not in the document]
RussTorres: If you are talking about the post_load, I could not find one with this full functionality.  in addition, the other schemas require a full render field and just allow it to supercede the RenderWebServiceParameters.  Again, it would be good to have a single superclass which could define this for client jobs, but I think that's outside the scope of this feature branch.

rendermodules/materialize/schemas.py
fcollman: Okay.. lets make an issue for this and address it later, could consolidate and reduce codebase.

rendermodules/materialize/schemas.py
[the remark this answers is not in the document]
fcollman: fair enough.. we should maybe add an issue to unify the RenderWebServiceParameters class and reutilize it across the board.

rendermodules/materialize/materialize_sections.py
fcollman: I think that makes sense for a location.

PR #73, opened 2018-01-31
Overwrite z option

RussTorres:
adding option (defaulting to false) to overwrite z layers in the output stacks by deleting those zs prior to upload.  This allows for more meaningful reprocessing in a workflow application where many operations have the same source and sink stacks.

### DIFF
--- a/integration_tests/test_consolidate.py
+++ b/integration_tests/test_consolidate.py
@@ -46,9 +46,10 @@ def test_consolidate_module(render,test_stack):
         "pool_size": 2,
         "minZ":input_z[1],
         "maxZ":np.max(input_z),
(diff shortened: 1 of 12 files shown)


fcollman:
I see a lot of the same code repeated here, should we consolidate it?

RussTorres:
I've been thinking about that because it was a little bit frustrating to handle the little differences in this that would break it.... some superclasses which handle input and output stacks along with the options they entail.
This is necessary for the workflow ASAP, so I'd prefer to get it into master before refactoring most of the modules.  maybe another simplification issue?

fcollman:
okay..sounds good. lets make an issue.

PR #77, opened 2018-02-03
Em montage qc

mahaling:
Adding clone stack feature to clone input montage stacks if they are in LOADING state


mahaling:
closing this to submit after integration tests

PR #79, opened 2018-02-07
apply rough alignment fixes

fcollman:
I changed code to catch the exceptions that were thrown in the function.  This revealed several bugs in the original tests that weren't passing correctly, including the set_new_z option, and some tests were failing due to incoherent json... not sure precisely why, but reengineering the code to use the import_tilespecs.  I also fixed the fact that the there was no output_json.  I also simplified some of the tilespec construction.


fcollman:
fixing up history in a new PR

PR #80, opened 2018-02-07
Rough align catch 2

fcollman:
I changed code to catch the exceptions that were thrown in the function. This revealed several bugs in the original tests that weren't passing correctly, including the set_new_z option, and some tests were failing due to incoherent json... not sure precisely why, but reengineering the code to use the import_tilespecs. I also fixed the fact that the there was no output_json. I also simplified some of the tilespec construction.

### DIFF
--- a/integration_tests/test_rough_align.py
+++ b/integration_tests/test_rough_align.py
@@ -211,6 +211,7 @@ def test_apply_rough_alignment_transform(render, montage_stack, test_do_rough_al
     ex1['minZ'] = 1020
     ex1['maxZ'] = 1022
     ex1['scale'] = 0.1
+    ex1['pool_size'] = 3
     ex1['output_json']=str(tmpdir_factory.mktemp('output').join('output.json'))
     ex1['loglevel']='DEBUG'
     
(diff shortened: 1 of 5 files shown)


fcollman:
I made some other changes to get over errors in the tilespecs... perhaps the lower pool size isn't necessary, i haven't experimented.

integration_tests/test_rough_align.py
```diff
@@ -210,6 +211,9 @@ def test_apply_rough_alignment_transform(render, montage_stack, test_do_rough_al
     ex1['minZ'] = 1020
     ex1['maxZ'] = 1022
     ex1['scale'] = 0.1
+    ex1['pool_size'] = 3
```
RussTorres: Should we set a test pool size (via environment variable) in `test_data` like we did for render-python?

fcollman: sure.. done

rendermodules/rough_align/schemas.py
```diff
@@ -425,3 +425,7 @@ def add_missing_values(self, data):
                 data['solver_options']['pastix']['ncpus'] = 8
             if pastix['split'] is None:
                 data['solver_options']['pastix']['split'] = 1
+
+class ApplyRoughAlignmentOutputParameters(DefaultSchema):
+    zs = argschema.fields.NumpyArray(description="list of z values that were applied to")
+    output_stack = argschema.fields.Str(descriptoin="stack where applied trasforms were set")
```
RussTorres: typo here in description.

mahaling:
I will run the module with real data to make sure everything works alright...

PR #82, opened 2018-02-08
Adding option to delete existing sections in target stack during montaging

mahaling:


mahaling:
closing it now as its missing some files from the last commit... will reopen after adding them

mahaling:
fixed issues with bamboo testing

integration_tests/test_em_montage.py
```diff
@@ -175,6 +175,10 @@ def test_run_montage_job_for_section(render,
     tilespecs = renderapi.tilespec.get_tile_specs_from_z(output_stack, solver_example['z_value'], render=render)
     assert len(tilespecs) == 4
 
+    # run this job again with same target collection to make sure that existing z in target stack gets deleted
+    mod = SolveMontageSectionModule(input_data=solver_example, args=[])
```
fcollman: how does this test for whether the existing z's are deleted? wouldn't we need to rerun it with fewer tiles and make sure that the output only had those tiles that were in it?

mahaling: Yes, that would be a more comprehensive test... Although this is just a preliminary version of how this case of existing sections should be handled. 
For example, in case of reimaging the existing section must not be deleted until the reimaged one passes montage qc. Or the other scenario would be to move this z to a new stack and then delete it later - again after qc gets passed... 

Hence, I am not elaborating this test as of now until we have a strategy planned out for reimaged sections.

mahaling: This is something that Tim and Russel wanted to have before we decide on the strategy...

PR #83, opened 2018-02-09
Faster seam detection and temp stack deletion 

mahaling:
Tested the code with Tim's stacks and improved seam detection code to run faster.


rendermodules/em_montage_qc/detect_montage_defects.py
```diff
@@ -224,10 +268,20 @@ def run(self):
...
+        # check if pre or post stitched stack is in LOADING state.
+        # if so, clone them to new stacks
+        status1, new_prestitched = check_status_of_stack(self.render,
+                                                self.args['prestitched_stack'],
+                                                zvalues)
+        
+        status2, new_poststitched = check_status_of_stack(self.render,
+                                                 self.args['poststitched_stack'],
+                                                 zvalues)
+        print(new_prestitched, new_poststitched)
```
fcollman: should be a self.logger.debug()

PR #84, opened 2018-02-10
apply rough alignment test refinement

fcollman:
I made some refinements to the tests to enable checking that we would catch an obvious error in the rough alignment.


fcollman:
going to make a cleaner history PR

PR #87, opened 2018-02-13
Point match optimization

mahaling:
point match optimization module. This module run the java point match client on the two tile images (given as input) on all possible combinations of parameters given and generates an html page showing all the point matches found for each parameter set.


integration_tests/test_point_match_optimization.py
```diff
@@ -0,0 +1,170 @@
...
+                        ptmatch_input_stack, render=render)))) == 0
+
+    yield ptmatch_input_stack
+    renderapi.stack.delete_stack(
+                        ptmatch_input_stack,
+                        render=render)
+    
+
+def test_pt_match_optimization_module(render, pt_match_input_stack, tmpdir_factory):
+    output_dir = str(str(tmpdir_factory.mktemp('pt_match_optimization')))
```
fcollman: why double str?

rendermodules/point_match_optimization/point_match_optimization.py
```diff
@@ -0,0 +1,313 @@
...
+                            qGroupId,
+                            self.args['outputDirectory'],
+                            self.args['url_options'],
+                            keys)
+
+        with renderapi.client.WithPool(self.args['pool_size']) as pool:
+            return_struct = pool.map(mypartial, options)
+
+        #print(os.path.dirname(__file__))
+        m = render_from_template(os.path.dirname(__file__), 'optimization.html', return_struct=return_struct)
```
RussTorres: This html should be included in the setup.py for distribution, I think in package_data?

rendermodules/point_match_optimization/point_match_optimization.py
```diff
@@ -0,0 +1,313 @@
...
+    return_struct['match_img_filename'] = match_img_filename
+    return_struct['collection_name'] = collection_name
+    return_struct['ptmatch_count'] = ptmatch_count
+    return_struct['tilepair_url'] = tilepair_base_url
+    return_struct['outdir'] = outdir
+    return return_struct
+
+
+class PointMatchOptimizationModule(RenderModule):
+    def __init__(self, schema_type=None, *args, **kwargs):
```
RussTorres: Do we want an output schema for this?

rendermodules/point_match_optimization/point_match_optimization.py
```diff
@@ -0,0 +1,313 @@
...
+    baseDataUrl = renderapi.render.format_baseurl(render.DEFAULT_KWARGS['host'], render.DEFAULT_KWARGS['port'])
+    argvs += ['--baseDataUrl', baseDataUrl]
+    argvs += ['--owner', render.DEFAULT_KWARGS['owner']]
+    argvs += ['--collection', collection_name]
+    #argvs += ['--matchStorageFile', os.path.join(outdir, 'matches.json')]
+    argvs += ['--debugDirectory', outdir]
+    argvs += ['', canvas_urls]
+
+    # run the point match client
+    renderapi.client.call_run_ws_client('org.janelia.render.client.PointMatchClient',
```
RussTorres: We should add this to render-python's client module.

rendermodules/point_match_optimization/point_match_optimization.py
```diff
@@ -0,0 +1,313 @@
...
+    tile_base_url = "%s/owner/%s/project/%s/stack/%s/tile"%(base_data_url,
+                                                            render_args['owner'],
+                                                            render_args['project'],
+                                                            stack)
+    url_suffix = "render-parameters"
+    if url_options['renderWithFilter'] is False:
+        url_suffix += '?filter=false'
+    else:
+        url_suffix += '?filter=true'
+    if url_options['normalizeForMatching'] is False:
```
fcollman: I think we should switch to using the newer transform labels and removing transform labels, rather than normalizeForMatching=true.  If in this case we want to excludeAllTransforms=true, or excludeTransformsAfterLast

rendermodules/point_match_optimization/point_match_optimization.py
fcollman: agreed. lets make a render-python issue and we can come back and replace this when that feature is done.

fcollman: i've made an issue

rendermodules/point_match_optimization/point_match_optimization.py
fcollman: good catch russel.. yes this will fail when we deploy it on the cluster given the way that our deployment process works now.

rendermodules/point_match_optimization/point_match_optimization.py
RussTorres: Agreed, but in this code it should be in addition to the normalizeForMatching functionality.  I think in the current implementation spark-based SIFT matching will assume normalizeForMatching=True.

fcollman: The current spark sift client runs on a tilepair file that contains the render-parameters template url, if alternate keyword arguments are given then the tilespecs returned will work.. what it assumes is that the tilespecs will be altered such that the image will be around 0,0 and the point matches will be extracted from those images as 'local' coordinates.

fcollman: that said i have no argument leaving it as an option.  We should also double back to the apply lens correction module and make sure there is an option to label the lens correction transform with a transform label, with a default of 'lens'.

fcollman: ok.. i think we won't add this for this release and we can come back later when the lens labels are included and we have fixed the spark client.

rendermodules/point_match_optimization/point_match_optimization.py
fcollman: I think this is addressed now

Issue #74, opened 2018-01-31
Shared classes for module patterns (input/output stacks)


RussTorres:
mentioned in #73, there is repeated code -- as well as code with the same function but different nomenclature -- throughout this repo.  Having a superclass which patterns input/output stacks and common options (close_stack, overwrite_z) would simplify processing.

fcollman:
closed via #88

Issue #71, opened 2018-01-31
Consolidate schemas..


fcollman:
We have a couple example of schemas.. namely RenderWebParameters and Spark parameters that are duplicated across several modules.  We should unify around a single version of that schema that is reused to make documentation and code changes in the future more easy.

fcollman:
closed via #88

PR #88, opened 2018-02-14
Simplify schemas

RussTorres:
refactor addressing #74 and #71

Adds schemas and argschemaparser classes to rendermodules.module which facilitate common stack operations and spark use cases to allow for code reuse.

Also important -- most tests now create new dictionary objects via copy or dict() prior to running tests as pre_load operations in schemas were changing the dictionary objects being used by other tests.

### DIFF
--- a/rendermodules/dataimport/apply_mipmaps_to_render.py
+++ b/rendermodules/dataimport/apply_mipmaps_to_render.py
@@ -69,13 +69,7 @@ class AddMipMapsToStack(StackTransitionModule):
 
     def run(self):
         self.logger.debug('Applying mipmaps to stack')
-
-        # get the list of z indices
-        zvalues = self.render.run(
-            renderapi.stack.get_z_values_for_stack, self.args['input_stack'])
-
-        zvalues = list(set(self.zValues).intersection(set(zvalues)))
-
+        zvalues = self.get_overlapping_inputstack_zvalues()
         if len(zvalues) == 0:
             self.logger.error('No sections found for stack {}'.format(
                 self.args['input_stack']))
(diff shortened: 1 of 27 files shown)


rendermodules/module/schemas/stack_schemas.py
```diff
@@ -0,0 +1,130 @@
...
+    maxZ = argschema.fields.Int(required=False)
+    z = argschema.fields.Int(required=False)
+    zValues = argschema.fields.List(argschema.fields.Int, required=False)
+
+    @post_load
+    def generate_zValues(self, data):
+        return self._generate_zValues(data)
+
+    def _generate_zValues(self, data):
+        print data
```
mahaling: this print statement should be removed

RussTorres: Good catch, fixed along with another one I left in.

rendermodules/dataimport/apply_mipmaps_to_render.py
```diff
@@ -74,15 +74,7 @@ def run(self):
         zvalues = self.render.run(
```
fcollman: I think this is a pretty common functionality that we might implement in the StackTransitionModule or a subclass of it.

rendermodules/module/schemas/stack_schemas.py
```diff
@@ -0,0 +1,129 @@
...
+        if 'zValues' in data:
+            return
+        elif 'z' in data:
+            data['zValues'] = [data['z']]
+        elif ('minZ' in data) and ('maxZ' in data):
+            data['zValues'] = range(data['minZ'], data['maxZ'] + 1)
+        else:
+            raise ValidationError("no z values specified")
+
+    def _override_input(self, data):
```
fcollman: can you add a comment with #DEPRECATED to help us find this when we want to get rid of it at some later point.

PR #96, opened 2018-04-03
Rough align fix

mahaling:
1. Fixing rough alignment code for translation factor in applying rough alignment. 
2. Adding parameters to map montage z to a new z value.


integration_tests/test_rough_align.py
```diff
@@ -262,11 +284,15 @@ def test_apply_rough_alignment_transform(render, montage_stack, test_do_rough_al
         mod.run()
 
     mod = ApplyRoughAlignmentTransform(input_data=ex3, args=[])
-    mod.run()
+    with pytest.raises(RenderModuleException):
```
RussTorres: I don't see what this is testing -- I think this `RenderModuleException` is actually a result of the base example filling in the new z value of 251, which causes a `RenderError` in getting tilespecs and is caught as a `RenderModuleException` when validating that the processes completed successfully.

RussTorres: Also, it would probably make sense to test the `map_z_start=True` capabilities in something other than the error case.

rendermodules/rough_align/apply_rough_alignment_to_montages.py
```diff
@@ -27,7 +27,8 @@
     "lowres_stack": "rough_test_downsample_rough_stack",
     "output_stack": "rough_test_rough_stack",
     "tilespec_directory": "/allen/programs/celltypes/workgroups/em-connectomics/gayathrim/nc-em2/Janelia_Pipeline/scratch/rough/jsonFiles",
-    "set_new_z": "False",
+    "map_z": "False",
+    "map_z_start": 251,
     "consolidate_trasnforms": "True",
```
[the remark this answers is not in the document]
mahaling: fixed and added more tests to check the behavior of code through mapping to a new z.

PR #97, opened 2018-04-06
Pathos cut

fcollman:
This fixes problems that are created by upgrading render-python to a version that doesn't use pathos for multiprocessing. I've also added a small test that is designed to see if with our present testing infrastructure that I get code coverage on simple partial/map patterns.  I'll take this out before we merge but I wanted to get the test in first to work out the problems.


fcollman:
woot! look at that code coverage now!!!

Issue #35, opened 2017-10-06
Non nearest-neighbor filtering for MipMapping


RussTorres:
Currently the only working implementation for creating mipmaps is a `PIL.resize()` which uses a nearest-neighbor resampling filter.  Additional options for this filter or different methods (there is a placeholder for `skimage.measure.block_reduce`-based area downsampling) will allow downsampling tasks to fit diverse workflows.

RussTorres:
Bumping per Stephan's suggestion that Nearest Neighbor screws everything up.

PR #103, opened 2018-04-25
Mipmap options

RussTorres:
Addresses #35 by introducing skimage block_reduce area downsampling.  As a bonus, includes filter options for the PIL method.


rendermodules/dataimport/create_mipmaps.py
```diff
@@ -26,6 +107,12 @@ def create_mipmaps(inputImage, outputDirectory='.', mipmaplevels=[1, 2, 3],
         whether to convert the image to 8 bit, dividing each value by 255
     force_redo: boolean
         whether to recreate mip map images if they already exist
+    method: str
+        string corresponding to downsampling method
```
fcollman: we should document the available methods here

fcollman:
my final comment would be that we need to test that block_reduce does the 'right' thing from render's perspective.

rendermodules/dataimport/schemas.py
```diff
@@ -18,13 +18,12 @@ class GenerateMipMapsParameters(InputStackParameters):
         description='directory to which the mipmaps will be stored')
     method = mm.fields.Str(
         required=True, default="block_reduce",
```
fcollman: this default never made sense before :(

Issue #99, opened 2018-04-16
Lens Correction Calculation is missing threaded SIFT


RussTorres:
a beanshell line like:
```java
param.maxNumThreadsSift = Integer.parseInt(System.getProperty("threads_SIFT"));
```

Would allow the lens correction script to utilize multiple threads.

PR #104, opened 2018-04-25
lens_correction: allow multithreaded SIFT

RussTorres:
introduces multithreaded SIFT matching implementation for LC addressing #99

### DIFF
--- a/rendermodules/lens_correction/schemas.py
+++ b/rendermodules/lens_correction/schemas.py
@@ -154,6 +154,9 @@ class LensCorrectionParameters(ArgSchema):
                      "unspecified or create temporary directory if None)"))
     SIFT_params = Nested(SIFTParameters)
     align_params = Nested(AlignmentParameters)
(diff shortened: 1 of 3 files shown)


rendermodules/lens_correction/schemas.py
```diff
@@ -154,6 +154,9 @@ class LensCorrectionParameters(ArgSchema):
                      "unspecified or create temporary directory if None)"))
     SIFT_params = Nested(SIFTParameters)
     align_params = Nested(AlignmentParameters)
+    max_threads_SIFT = Int(
+        required=False, default=1, description=(
```
fcollman: is there a reason to have this as 1 as a default?

RussTorres: Not really -- I prefer to have it low, but if you think it's useful to to have it default to more we can do that.

fcollman: How about 3?

PR #105, opened 2018-04-25
Docs update

fcollman:


docs_requirements.txt
```diff
@@ -1,4 +1,17 @@
...
+pillow
+tifffile
+pathlib
+scipy
+rtree
+networkx
+bokeh
+matplotlib
+opencv-python
+jinja2
```
RussTorres: should this include scikit-image after #103?

fcollman:
yes let me make sure that this fixes the rtd build before i reopen this

fcollman:
ok i have this working now
http://render-modules.readthedocs.io/en/docs_update/

yes, we need to add to this docs_requirements file when we updated requirements.. we can have it point to our normal requirments, but then we need to put all the doc specific requirements in there.  readthedocs won't allow me to use 2 different pip requirements files for setting up the virtualenv, and opencv-python won't setuptools install.. it will only pip install.  Another reason to get rid of it if we can.

PR #113, opened 2018-06-07
updated point match optimization

mahaling:
1. Reads in a tile pair file and randomly picks the specified number of tilepairs for pt match optimization
2. Includes a filter to include only those tile pairs that actually overlap for pt match optimization
3. Output is a html file that has the matches and parameters tabulated
4. Changed some functions to avoid race conditions in parallel processing


rendermodules/point_match_optimization/pt_match_optimization.py
```diff
@@ -0,0 +1,583 @@
...
+import shutil
+import tempfile
+import time
+import urllib
+import matplotlib.pyplot as plt
+import matplotlib.image as mpimg
+from numbers import Number
+from functools import partial
+from operator import itemgetter
+from shapely.geometry import Polygon
```
RussTorres: We should add shapely to requirements.txt

rendermodules/point_match_optimization/pt_match_optimization.py
```diff
@@ -0,0 +1,583 @@
...
+
+def render_from_template(directory, template_name, **kwargs):
+    loader = FileSystemLoader(directory)
+    env = Environment(loader=loader)
+    template = env.get_template(template_name)
+    return template.render(**kwargs)
+
+def draw_matches(render, stack, tile1, tile2, ptmatches, scale, outdir, color=None):
+    
+    im1_url = "%s:%d/render-ws/v1/owner/%s/project/%s/stack/%s/tile/%s/png-image?scale=%f&filter=false&normalizeForMatchin=true"%(render.DEFAULT_HOST,
```
RussTorres: small typo in normalizeForMatching here and below -- but it seems like these should also match the url params given in the input json rather than a default.

rendermodules/point_match_optimization/pt_match_optimization.py
```diff
@@ -0,0 +1,583 @@
...
+
+    print(im1_url)
+    print(im2_url)
+    r = urllib.urlopen(im1_url)
+    img1 = np.asarray(bytearray(r.read()), dtype="uint8")
+    img1 = cv2.imdecode(img1, cv2.IMREAD_COLOR)
+
+    r = urllib.urlopen(im2_url)
+    img2 = np.asarray(bytearray(r.read()), dtype="uint8")
+    img2 = cv2.imdecode(img2, cv2.IMREAD_COLOR)
```
RussTorres: Don't think it's detrimental, but you should be able to use `renderapi.image.get_tile_image_data` to do something like this as well.

rendermodules/point_match_optimization/pt_match_optimization.py
fcollman: agreed

rendermodules/point_match_optimization/pt_match_optimization.py
```diff
@@ -0,0 +1,583 @@
...
+
+
+'''
+def get_canvas_url(render_args, stack, tile1, tile2, url_options):
+    base_data_url = '%s:%d/render-ws/v1'%(render_args['host'], render_args['port'])
+    tile_base_url = "%s/owner/%s/project/%s/stack/%s/tile"%(base_data_url,
+                                                            render_args['owner'],
+                                                            render_args['project'],
+                                                            stack)
+    url_suffix = "render-parameters"
```
fcollman: These are not all the urlOptions that are available.  I think either we should list them all here explicitly or have a wrapper function for converting them to query arguments in a standard fashion and accept them all.  Noteably, its missing handling of the excludeAllTransforms, excludeFirstTransformAndAllAfter, excludeTransformsAfterLast.

fcollman: These options are still not wired up in this part of the code.. I since we schematized them we should just have a input as a dictionary of query parameters arguments and append them all to the url, rather than listed them out explicitly, if someone passes an invalid query argument to this function it won't break anything, it just won't complain.  We could re-validate the dictionary with the schema in this function, but I think that's overkill.

mahaling: changes have been made

mahaling: removed it

PR #115, opened 2018-06-12
Rough align z mapping

mahaling:
changing the input z range in applying rough alignment to montages into a list of old z values and their mapping z values.

### DIFF
--- a/rendermodules/rough_align/apply_rough_alignment_to_montages.py
+++ b/rendermodules/rough_align/apply_rough_alignment_to_montages.py
@@ -201,20 +201,9 @@ class ApplyRoughAlignmentTransform(RenderModule):
                             self.args['output_stack'],
                             'LOADING')
 
-        #with renderapi.client.WithPool(self.args['pool_size']) as pool:
-        #    results=pool.map(mypartial, Z)
-        results = []
-        for zz in Z:
-            results.append(apply_rough_alignment(self.render,
-                                                 self.args['montage_stack'],
-                                                 self.args['prealigned_stack'],
-                                                 self.args['lowres_stack'],
-                                                 self.args['output_stack'],
-                                                 self.args['tilespec_directory'],
-                                                 self.args['scale'],
-                                                 zz,
-                                                 consolidateTransforms=self.args['consolidate_transforms']))
-
+        with renderapi.client.WithPool(self.args['pool_size']) as pool:
+            results=pool.map(mypartial, Z)
+        
         #raise an exception if all the z values to apply alignment were not
         if not all([r is None for r in results]):
             failed_zs = [(result,z) for result,z in zip(results,Z) if result is not None]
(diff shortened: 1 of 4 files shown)


rendermodules/rough_align/schemas.py
```diff
@@ -50,22 +56,41 @@ class ApplyRoughAlignmentTransformParameters(RenderParameters):
...
+    
 
     @post_load
     def validate_data(self, data):
         if data['prealigned_stack'] is None:
             data['prealigned_stack'] = data['montage_stack']
         if data['map_z']:
-            if data['map_z_start'] == -1:
-                raise ValidationError("map_z_start is invalid")
+            if data['new_z'][0] == -1:
```
RussTorres: Is there a reason to have this default to `[-1]`?  `None` or missing might be more transparent for use cases which have negative numbers.

rendermodules/rough_align/apply_rough_alignment_to_montages.py
```diff
@@ -151,10 +137,11 @@ def apply_rough_alignment(render,
             t.z = newz
             #t.z = z
             t.layout.sectionId = "%s.0"%str(int(newz))
-
+        
         renderapi.client.import_tilespecs(
             output_stack, highres_ts1,
             sharedTransforms=sharedTransforms_highrests1, render=render)
+        print('getting tilespecs from')
```
fcollman: should probably nuke this print statement or convert it to a logging statement

rendermodules/rough_align/schemas.py
fcollman: perhaps it was to avoid exceptions when trying to access those keys? I could suggest this instead..

```
if data['map_z']:
    try:
        if abs(len(data['new_z']) - len(data['old_z'])) != 0:
            raise ValidationError("new_z list count does not match with old_z list count")
    except KeyError:
            raise ValidationError("new_z and old_z must be specified if map_z is True")
```

rendermodules/rough_align/apply_rough_alignment_to_montages.py
```diff
@@ -164,28 +151,32 @@ def apply_rough_alignment(render,
 class ApplyRoughAlignmentTransform(RenderModule):
     default_schema = ApplyRoughAlignmentTransformParameters
     default_output_schema = ApplyRoughAlignmentOutputParameters
+    
     def run(self):
-        if self.args['minZ'] > self.args['maxZ']:
-            raise RenderModuleException("First section z cannot be greater or equal to last section z")
+        #if self.args['minZ'] > self.args['maxZ']:
```
fcollman: I think we should delete the old code in the comments.. we have it in version control so we can get it back easily.

rendermodules/rough_align/schemas.py
```diff
@@ -50,22 +56,41 @@ class ApplyRoughAlignmentTransformParameters(RenderParameters):
...
+                raise ValidationError("new_z is invalid. You need to specify new_z as a list of values")
+            elif abs(len(data['new_z']) - len(data['old_z'])) != 0:
+                raise ValidationError("new_z list count does not match with old_z list count")
         else:
-            data['map_z_start'] = data['minZ']
+            data['new_z'] = data['old_z']
+
+
+    #@post_load
```
fcollman: same comment about comments here.

PR #116, opened 2018-06-13
updating plots in em montage qc

mahaling:
adding the latest version of the plot that now shows the defects in the plot

### DIFF
--- a/rendermodules/em_montage_qc/detect_montage_defects.py
+++ b/rendermodules/em_montage_qc/detect_montage_defects.py
@@ -31,28 +31,6 @@ example = {
     "pool_size": 20
 }
 
-'''
-example = {
-    "render":{
-        "host": "http://em-131db",
-        "port": 8080,
-        "owner": "TEM",
-        "project": "247488_8R",
-        "client_scripts": "/allen/programs/celltypes/workgroups/em-connectomics/gayathrim/nc-em2/Janelia_Pipeline/render_latest/render-ws-java-client/src/main/scripts"
-    },
-    "prestitched_stack": "em_2d_montage_lc",
-    "poststitched_stack": "em_2d_montage_solved_TEMCA2_1222_1260",
-    "match_collection_owner": "TEM",
-    "match_collection": "default_point_matches",
-    "minZ": 1242,
-    "maxZ": 1244,
-    "residual_threshold":6,
-    "neighbor_distance": 50,
-    "min_cluster_size":20,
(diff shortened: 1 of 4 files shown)


rendermodules/em_montage_qc/detect_montage_defects.py
```diff
@@ -362,9 +362,12 @@ def run(self):
         centroids = [seam_centroids[i] for i in seams_indices]
 
         self.args['output_html'] = self.args['out_html_dir']
-        if len(combinedz) > 0:
-            if self.args['plot_sections']:
-                self.args['output_html'] = plot_section_maps(self.render, new_poststitched, combinedz, out_html_dir=self.args['output_html'])
+        if self.args['plot_sections']:
+            self.args['output_html'] = plot_section_maps(self.render, self.args['poststitched_stack'], post_tspecs, disconnected_tiles, gap_tiles, seam_centroids, zvalues, out_html_dir=self.args['output_html'])
```
RussTorres: I'm confused by the use of the `output_html` and `out_html_dir` here -- the data types change a lot in these few lines and ultimately end up invalidating the schema -- are these used somewhere where they need to be attached to args?

rendermodules/em_montage_qc/schemas.py
```diff
@@ -54,7 +54,8 @@ def add_match_collection_owner(self, data):
 
 
 class DetectMontageDefectsParametersOutput(argschema.schemas.DefaultSchema):
-    output_html = Str(
+    output_html = argschema.fields.List(
```
RussTorres: Does `output_html` still work?  I don't see where this is used as a list.

rendermodules/em_montage_qc/detect_montage_defects.py
mahaling: changing this to avoid confusion in the schema... the older version of which had out_html_dir and output_html as Strings

rendermodules/em_montage_qc/schemas.py
mahaling: output_html is used by blue_sky to link the plot to the web interface showing status of each montage set

mahaling: this would be a list if minZ and maxZ is a range and not hold the same value.

RussTorres: I misread this -- Thought it was in the input schema and now it makes a lot more sense.

PR #118, opened 2018-06-29
Mesh lc tests

mahaling:
Mesh based lens correction... 

working version with some validation to adapt to workflow. 

have added only basic integration tests. not complicated.

### DIFF
--- a/integration_tests/test_mesh_lens_correction.py
+++ b/integration_tests/test_mesh_lens_correction.py
@@ -64,9 +64,6 @@ def test_mesh_lens_correction(render, example_tform_dict, tmpdir_factory):
     example['out_html_dir'] = out_html_dir
     example['z_index'] = 100
 
-    outfile = tmpdir_factory.mktemp("mesh_lens").join("lens_output.json")
-    example['outfile'] = outfile
-
     mod = MeshLensCorrection(input_data=example, args=['--output_json', 'mesh_lens_out.json'])
     mod.run()
     


rendermodules/mesh_lens_correction/LensPointMatches.py
```diff
@@ -0,0 +1,296 @@
...
+        ncpus = multiprocessing.cpu_count()
+    
+    args = {}
+    args['downsample_scale'] = downsample_scale
+    args['render'] = render 
+    args['match_collection'] = match_collection
+    args['nfeature_limit'] = nfeature_limit
+    args['matchMax'] = matchMax
+
+    pool = multiprocessing.Pool(ncpus)
```
RussTorres: Using `WithPool` defined in the renderapi client should make these calls testable without mocking non-forked processes.

rendermodules/mesh_lens_correction/LensPointMatches.py
```diff
@@ -0,0 +1,296 @@
...
+    
+    fargs = []
+
+    if index_list is None:
+        index_list = range(tile_index.shape[0])
+    
+    for i in index_list:
+        impaths = [t.ip.mipMapLevels[0].imageUrl.split('file://')[-1]
```
RussTorres: Probably safer here to use the file uri handler in the form of `urllib.unquote(urlparse.urlparse(str(oldUrl)).path)`.

integration_tests/test_mesh_lens_correction.py
```diff
@@ -0,0 +1,98 @@
...
+    with open(js['output_json'], 'r') as f:
+        new_tform_dict = json.load(f)
+
+    with open(js['qc_json'], 'r') as f:
+        qc = json.load(f)
+
+    assert(not qc['hole_sections'] and not qc['gap_sections'])
+
+    #example_tform = renderapi.transform.ThinPlateSplineTransform(dataString=example_tform_dict['dataString'])
+    #new_tform = renderapi.transform.ThinPlateSplineTransform(dataString=new_tform_dict['dataString'])
```
RussTorres: How difficult is it to add a residual test comparison based on e.g. a grid?

rendermodules/mesh_lens_correction/LensPointMatches.py
mahaling: Yes.. agreed but this and other changes will come in the next version of refactoring without changing the logic of the code... This PR is to just facilitate ease of processing for Tim.

integration_tests/test_mesh_lens_correction.py
mahaling: yes planning to add this in next PR

RussTorres: I added issue #119 as a reminder so I feel okay about merging this.

PR #128, opened 2018-07-07
requirements: bump em_aligner to support overwrite zvalues

RussTorres:
bumps em aligner version to support overwrite z capability


djkapner:
Same as #127?

RussTorres:
@djkapner so it is....

PR #130, opened 2018-07-10
Add coarse mesh failure

djkapner:
quality of life Improvements in MeshAndSolveTransform.
No similar improvements for point matching yet.
Initial PR for checking integration tests.

### DIFF
--- a/integration_tests/test_mesh_lens_correction.py
+++ b/integration_tests/test_mesh_lens_correction.py
@@ -41,6 +41,26 @@ def render():
     return render
 
 
+def test_coarse_mesh_failure(render, tmpdir_factory):
+    outdir = str(tmpdir_factory.mktemp("mesh_lens"))
+    out_html_dir = outdir
+    example['render'] = render_params
+    example['metafile'] = os.path.join(TEST_DATA_ROOT,
+                                       "em_modules_test_data",
+                                       "mesh_lens_correction_3",
+                                       "mesh_lens_metafile_3.json")
+    example['output_dir'] = outdir
+    example['out_html_dir'] = out_html_dir
+    example['z_index'] = 101
+
+    mod = MeshLensCorrection(
+            input_data=example,
+            args=['--output_json', 'mesh_lens_out.json'])
+
+    with pytest.raises(AssertionError):
+        mod.run()
+
+
 def test_mesh_lens_correction(render, tmpdir_factory):
     outdir = str(tmpdir_factory.mktemp("mesh_lens"))
     out_html_dir = outdir
--- a/integration_tests/test_mesh_lens_correction.py
+++ b/integration_tests/test_mesh_lens_correction.py
@@ -91,7 +91,7 @@ def test_mesh_lens_correction(render, tmpdir_factory):
     renderapi.pointmatch.import_matches(
             example['match_collection'],
             [match],
-            render=renderapi.connect(**example['render']))
+            render=mod.render)
 
     mod.run()
 
--- a/integration_tests/test_mesh_lens_correction.py
+++ b/integration_tests/test_mesh_lens_correction.py
@@ -71,6 +71,7 @@ def test_mesh_lens_correction(render, tmpdir_factory):
                                        "mesh_lens_metafile_2.json")
     example['output_dir'] = outdir
     example['out_html_dir'] = out_html_dir
+    example['outfile'] = os.path.join(outdir,'out.json')
     example['z_index'] = 100
 
     mod = MeshLensCorrection(
--- a/integration_tests/test_mesh_lens_correction.py
+++ b/integration_tests/test_mesh_lens_correction.py
@@ -4,6 +4,7 @@ import renderapi
 import os
 
 from rendermodules.mesh_lens_correction.do_mesh_lens_correction import MeshLensCorrection
+from rendermodules.mesh_lens_correction.MeshAndSolveTransform import MeshLensCorrectionException
 from rendermodules.mesh_lens_correction.LensPointMatches import generate_point_matches
 from test_data import render_params, TEST_DATA_ROOT
 
@@ -57,7 +58,7 @@ def test_coarse_mesh_failure(render, tmpdir_factory):
             input_data=example,
             args=['--output_json', 'mesh_lens_out.json'])
 
-    with pytest.raises(AssertionError):
+    with pytest.raises(MeshLensCorrectionException):
         mod.run()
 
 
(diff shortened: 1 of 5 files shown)


rendermodules/mesh_lens_correction/MeshAndSolveTransform.py
```diff
@@ -539,88 +408,154 @@ def create_A(self):
...
+            transforms.append(renderapi.transform.AffineModel(
+                               M00=solution[0][i*3+0],
+                               M01=solution[0][i*3+1],
+                               B0=solution[0][i*3+2],
+                               M10=solution[1][i*3+0],
+                               M11=solution[1][i*3+1],
+                               B1=solution[1][i*3+2]))
+        return transforms
+
+    def MeshAndSolve(self):
```
fcollman: this is bad python style... CapitalLetterCase indicates this is a class that can be instantiated.

djkapner:
Was run(). Can change back

On Jul 9, 2018 7:49 PM, Forrest Collman <notifications@github.com> wrote:

@fcollman commented on this pull request.

________________________________

In rendermodules/mesh_lens_correction/MeshAndSolveTransform.py<https://na01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FAllenInstitute%2Frender-modules%2Fpull%2F130%23discussion_r201204125&data=02%7C01%7C%7C26b250f45a61404753fc08d5e60fc72c%7C32669cd6737f4b398bddd6951120d3fc%7C0%7C0%7C636667877801939019&sdata=Wql7N%2BRxupd6QWucbM2LKNvVtyGAUGIWSB6240jtIsU%3D&reserved=0>:

> +        wts.data = weights
+        return A, wts, lens_dof_start
+
+    def create_transforms(self, ntiles, solution):
+        transforms = []
+        for i in range(ntiles):
+            transforms.append(renderapi.transform.AffineModel(
+                               M00=solution[0][i*3+0],
+                               M01=solution[0][i*3+1],
+                               B0=solution[0][i*3+2],
+                               M10=solution[1][i*3+0],
+                               M11=solution[1][i*3+1],
+                               B1=solution[1][i*3+2]))
+        return transforms
+
+    def MeshAndSolve(self):


this is bad python style... CapitalLetterCase indicates this is a class that can be instantiated.

djkapner:
static methods or the outside class sound fine to me. Would like feedback from @mahaling and @RussTorres as well before I keep going, so everyone is happy with how it is.

mahaling:
Hey Dan... I agree with Forrest and that's what I meant in our discussion earlier today in moving those functions out of the class and thus passing explicit arguments to each one of them.

integration_tests/test_mesh_lens_correction.py
```diff
@@ -41,6 +41,26 @@ def render():
     return render
 
 
+def test_coarse_mesh_failure(render, tmpdir_factory):
+    outdir = str(tmpdir_factory.mktemp("mesh_lens"))
+    out_html_dir = outdir
+    example['render'] = render_params
```
RussTorres: We have been using something like
```python
example_for_input = dict(example, **{"render": render_params})
```
in our tests since this makes a copy of the example data and does not risk contaminating other tests in the case of changes via schema validation or assignments in the module.

rendermodules/mesh_lens_correction/MeshAndSolveTransform.py
```diff
@@ -11,616 +11,569 @@
...
+                    (coords[:, 1] >= c.min()) &
+                    (coords[:, 1] <= c.max())).flatten()
+            a = np.arange(ind.size)
+            np.random.shuffle(a)
+            ind = ind[a[0:min_count]]
+            new_coords.append(coords[ind])
+    return np.concatenate(new_coords)
+
+
+def write_montage_qc_json(args, tilespecs):
```
RussTorres: This function seems unnecessary -- could this just be written as a dictionary and output after validating an output schema?

rendermodules/mesh_lens_correction/do_mesh_lens_correction.py
```diff
@@ -1,25 +1,30 @@
...
 
 example = {
     "render": {
         "host": "em-131snb1",
         "port": 8080,
         "owner": "gayathrim",
         "project": "lens_corr",
-        "client_scripts":"/allen/aibs/pipeline/image_processing/volume_assembly/render-jars/production/scripts",
-        "memGB":"2G"
+        "client_sxripts": "/allen/aibs/pipeline/image_processing/volume_assembly/render-jars/production/scripts",
```
gayathrimahalingam: typo

rendermodules/mesh_lens_correction/do_mesh_lens_correction.py
```diff
@@ -143,46 +155,47 @@ def run(self):
...
-        # generate point matches
-        generate_point_matches(self.render,
-                               self.args['pairJson'],
-                               self.args['input_stack'],
-                               self.args['match_collection'],
-                               self.args['matchMax'],
-                               nfeature_limit=self.args['nfeature_limit'])
-
-        meshclass = MeshAndSolveTransform()
+        if self.args['rerun_pointmatch']:
```
gayathrimahalingam: shouldn't the existing point matches be deleted by default if they exists?. This will make the get_tilespecs_and_matches function in MeshAndSolve to fail if the tile_ids that exist in point matches don't exist in the stack.

The get_tilespecs_and_matches is getting its "unique_ids" from the match collection and not the tilespecs.

djkapner: for debugging purposes, especially when just working with the mesher/solver/transform maker, it is nice to not have to rerun the pointmatches.
it also helps with test coverage without having to rerun the pointmatches in every case.
the default is to delete them.

djkapner: also, the get_tilespecs_and_matches is getting unique ids from the intersection of stacks and matches.

gayathrimahalingam: ok makes sense

rendermodules/mesh_lens_correction/MeshAndSolveTransform.py
```diff
@@ -11,616 +11,584 @@
...
+        # and enforce neighboring matches to vertices
+        self.mesh, self.area_triangle_par = \
+            force_vertices_with_npoints(
+                self.area_triangle_par,
+                self.bbox,
+                self.coords,
+                3)
+
+        try:
+            assert self.mesh.points.shape[0] > 0.5*self.args['nvertex'], \
```
RussTorres: Assertions in python are always touchy -- running optimized python (e.g. with the `-O`  flag) disables them altogether which may or may not be what you want at various points here.

djkapner: you're right. I don't want them disabled. How would you suggest I assert or check numerical results? `if` then `raise`?

PR #131, opened 2018-07-10
make_montage_scape: reduce image directory requirements -- allow simp…

RussTorres:
…le generation

This had previously succeeded in the workflow with already existing directories, but fails on some newer cases.

### DIFF
--- a/rendermodules/dataimport/schemas.py
+++ b/rendermodules/dataimport/schemas.py
@@ -6,7 +6,6 @@ from marshmallow import ValidationError, post_load
 from argschema.schemas import DefaultSchema
 
 
-
 class GenerateMipMapsOutput(DefaultSchema):
     levels = Int(required=True)
     output_dir = Str(required=True)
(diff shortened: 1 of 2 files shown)


rendermodules/dataimport/schemas.py
```diff
@@ -108,7 +107,7 @@ class MakeMontageScapeSectionStackParameters(OutputStackParameters):
     montage_stack = Str(
         required=True,
         metadata={'description':'stack to make a downsample version of'})
-    image_directory = InputDir(
+    image_directory = Str(
```
fcollman: Shouldn't this be an OutputDir?

fcollman: no .. i'm wrong.  But now i'm still confused why InputDir is wrong?  This should prevent the module from running with incorrectly specified input directories? Is there an argschema bug?

RussTorres: @fcollman The point of this is that the module has the option of generating the downsample if it does not exist.  We are currently using the functionality in the workflow without using the downsampling strategy proper, so it is failing am a nonexistent `image_directory` even though the module would do what is intended.  `OutputDir` isn't quite right either since it requires the value be writable -- it could be read only if the montage has previously been generated.

fcollman: k... I get it now.  The directory needs to either already exist OR be writeable.  which is neither InputDir or OutputDir.

PR #135, opened 2018-07-12
upgraded EM aligner to 0.1.6

sharmishtaa:


djkapner:
Russel noticed that  sectionId = renderapi.stack.get_sectionId_for_z() is not going to work for LOADING stacks

djkapner:
looks like I need to tweak the requirements in EMaligner to have render-python==1.13.0 so it doesn't go up to 2.0 yet. I'll do that in about an hour.

fcollman:
Hi Dan... i think that EMAligner does not have this problem... it's other
modules.  Since your tests passed, it means you are fine with 2.0, if you
brought in 2.0 in the testing.  I think you probably are given that 2.0
changes are related to image pyramids that the solver doesn't interact with
at all.  The <2.0 flag should be on render-modules for now till we get that
up to speed.  If you tag EMAligner wtih 1.13.0 then there will eventually
be an incompatibliity between EMAligner and render-modules.

djkapner:
understood. I was trying to help get something in Sharmi's hands before I leave this afternoon (for 2 weeks). Along those lines, I would suggest cloning the current render production conda, pip uninstalling EMaligner v0.1.4 and pip installing v0.1.15. At least that will get you something to work with.

fcollman:
yeah i think to get something working today we should make a PR to render
modules fixing the render-python requirements.

i've  started  the renderpython2 branch.

RussTorres:
@sharmishtaa this should be 0.1.6 as 0.1.5 breaks the EM workflow.

fcollman:
i rebased this onto develop to get the render-python updates

PR #136, opened 2018-07-13
render-python 2.0 compatability

fcollman:
This PR is meant to update the modules to use the new imagepyramid interface defined in render-python 2.0.


rendermodules/dataimport/generate_EM_tilespecs_from_metafile.py
```diff
@@ -72,7 +72,7 @@ def ts_from_imgdata(self, imgdata, imgdir, x, y,
             minint=minint, maxint=maxint,
             tforms=raw_tforms,
             mipMapLevels=[
-                renderapi.tilespec.MipMapLevel(
+                renderapi.image_pyramid.MipMapLevel(
```
fcollman: I should probably change this to not use mipMapLevels.

run_git_log.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
fcollman: this file was probably added by mistake?

PR #137, opened 2018-07-14
Python3compat

RussTorres:
Most tests are confirmed to work with python3.6 as well as python2.7 -- python 3.7 is failing because opencv-contrib-python is not yet compatible, but hopefully that gets resolved soon.

Edit:
whoops, accidentally referenced opencv-contrib-python issue 106 associated with this and github won't let me unreference that.


integration_tests/test_consolidate.py
```diff
@@ -134,7 +133,11 @@ def test_redirect_mipMapLevels(render, test_stack, tmpdir):
     modified_tspecs = renderapi.tilespec.get_tile_specs_from_z(
         output_stack, z, render=render)
 
+<<<<<<< HEAD
```
fcollman: missed merge

integration_tests/test_rough_align.py
```diff
@@ -571,7 +570,11 @@ def test_make_montage_stack_module_without_downsamples(
     tspecs = render.run(
         renderapi.tilespec.get_tile_specs_from_stack, output_stack)
 
+<<<<<<< HEAD
```
fcollman: missed merge

RussTorres:
Also, by "most" test I mean all but the EM_aligner_python test which is problematic because I don't have a mongo port exposed for testing.  I think EM_aligner_python is all py3.7 compatible.

rendermodules/em_montage_qc/detect_montage_defects.py
```diff
@@ -146,7 +146,8 @@ def detect_stitching_gaps(render, prestitched_stack, poststitched_stack,
     gap_tiles = []
     for n in G2.nodes():
         if G1.degree(n) > G2.degree(n):
-            tileId = pre_tileIds.keys()[pre_tileIds.values().index(n)]
+            # FIXME is this correct for python2?
```
fcollman: I think this is okay...

RussTorres: Right , it looks like nothing happens to that dictionary after creation that would change the order.  Will remove comment,

PR #139, opened 2018-07-16
Mipmaptypefix

RussTorres:
block_reduce was previously not preserving the numpy array type and writing mipmaps in float mode 'F' rather than uint8 'L'.  I think Render is naive in interpreting this and did not notice, but this is a high-priority fix.

### DIFF
--- a/integration_tests/test_dataimport.py
+++ b/integration_tests/test_dataimport.py
@@ -49,12 +49,14 @@ def test_generate_EM_metadata(render):
     renderapi.stack.delete_stack(ex['stack'], render=render)
     assert len(expected_tileIds.symmetric_difference(delivered_tileIds)) == 0
 
-def validate_mipmap_generated(in_ts,out_ts,levels,imgformat='tif'):
+
+def validate_mipmap_generated(in_ts, out_ts, levels, imgformat='tif',
+                              targetmode='L', **kwargs):
     # make sure that the corresponding tiles' l0s match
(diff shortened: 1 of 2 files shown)


integration_tests/test_dataimport.py
```diff
@@ -67,10 +69,11 @@ def validate_mipmap_generated(in_ts,out_ts,levels,imgformat='tif'):
 
     for lvl, mmL in iteritems(out_ts.ip):
         fn = urllib.parse.unquote(urllib.parse.urlparse(mmL.imageUrl).path)
-        ext=os.path.splitext(fn)[1]
+        ext = os.path.splitext(fn)[1]
         if (lvl != "0"):
             assert ext[1:] == imgformat
         with Image.open(fn) as im:
+            assert im.mode == targetmode
```
fcollman: if we have this assert, why are we giving the option to set the targetmode with the default = 'L', shouldn't the default be to keep the mode of the input file, and if the targetmode is set, to set it to be the output mode.  If I recall correctly, render can handle correctly 16bit tiff mipmap 0 levels and 8bit jpeg mipmaps.

RussTorres: This is a good point -- I am not sure from where the code coverage on converting from 16 bit to 8 bit is coming, but this change doesn't seem to affect it.  I'm not too concerned, but that's something that should probably be fixed..

PR #141, opened 2018-07-17
rough_align: include lower level mipmaps in downsample tempstack

RussTorres:
Some extremely distorted tiles have been causing issues in materializing downsampled montages.  These cases are looking for level 0 mipmaps and failing with an error like:
```
java.lang.IllegalArgumentException: The highest resolution mipmap for file:/allen/programs/celltypes/production/wijem/workflow_data/production/mipmap/em_montage_17797_1R_z130208_2018_06_18_18_15_56_00_00/1/allen/programs/celltypes/production/wijem/incoming_data/17797_1R_Tape151_TEMCA2_04_000208_0/20180618111803520_17797_1R_Tape151_TEMCA2_04_000208_0_45_3.tif.tif is level 1 but a level 0 mipmap is needed.  Upscaling is not currently supported.
```

This PR modifies downsample generation to keep all tilespecs below the target level, allowing level 0s to be used when necessary.

### DIFF
--- a/integration_tests/test_rough_align.py
+++ b/integration_tests/test_rough_align.py
@@ -457,7 +457,14 @@ def test_render_downsample_with_mipmaps(render, one_tile_montage, tmpdir_factory
     ex3 = dict(ex2, **{'minZ': 1 , 'maxZ': 2})
 
     #stack_has_mipmaps = check_stack_for_mipmaps(render, ex['input_stack'], [1020])
-    tempjson = create_tilespecs_without_mipmaps(render, ex['input_stack'], 0, 1020)
+    # tempjson = create_tilespecs_without_mipmaps(render, ex['input_stack'], 0, 1020)
+
+    # generate tilespecs used for rendering stack input
+    maxlvl = 1
+    tspecs = create_tilespecs_without_mipmaps(render, ex['input_stack'], maxlvl, 1020)
+    ts_levels = {int(i) for l in (ts.ip.levels for ts in tspecs) for i in l}
+    # levels > maxlevel should not be included if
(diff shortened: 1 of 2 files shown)


fcollman:
Our solution in AT to a similar problem was to produce a dropped stack where we look for tiles which we believe are misstitched. Since the goal of rough alignment is a single transform per section, we can apply the answer back to a non dropped stack.

RussTorres:
That would be an acceptable solution, but this only happens in montages that have passed QC -- so we believe that they don't seriously affect the rendered montage quality.  In my mind this problem still constitutes a bug because of how Render treats level 0 mipmaps.

fcollman:
sorry i blanked on this.... I am not able to mentally follow the presence of mipmap 0 levels through the whole process... but a generic concern is that their presence will trigger their use more often than we think it technically should and slow things way down.  So have we done a trial run on the critical modules for performance with such stacks downstream?

RussTorres:
I think an alternative implementation that adds more functionality is to change (or add support for) a "levels" input rather than the "level" input used to generate the downsample.  We currently have all mipmaplevels available at all times, so I'm not concerned about something like a level 0 read failure (which had affected us in the past.)

FWIW, the current use case has changed to using even smaller downsampled sections and at the current scale this issue does not appear.

mahaling:
just curious... what is the scale parameter used for generating the downsample sections (in this particular example)?.

RussTorres:
@mahaling this was a failure mode when we were using 0.1 scale montages in the workflow, but is no longer hit with the current 0.01 scale parameters.

PR #146, opened 2018-08-02
parameterizing solve criteria, setting scale to 0.1 default

djkapner:


rendermodules/mesh_lens_correction/schemas.py
```diff
@@ -23,6 +23,24 @@ class regularization(ArgSchema):
         description="regularization for lens parameters")
 
 
+class good_solve_criteria(ArgSchema):
```
fcollman: this should be a DefaultSchema

PR #148, opened 2018-08-09
Roughalign scale option

mahaling:
adding the scale option in downsample montage stack and apply rough alignment modules as an option controlled by a boolean parameter

### DIFF
--- a/rendermodules/rough_align/apply_rough_alignment_to_montages.py
+++ b/rendermodules/rough_align/apply_rough_alignment_to_montages.py
@@ -173,6 +173,7 @@ class ApplyRoughAlignmentTransform(RenderModule):
                         self.args['output_stack'],
                         self.args['tilespec_directory'],
                         self.args['scale'],
+                        apply_scale=self.args['apply_scale'],
                         consolidateTransforms=self.args['consolidate_transforms'])
 
         # Create the output stack if it doesn't exist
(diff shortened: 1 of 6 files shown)


rendermodules/rough_align/apply_rough_alignment_to_montages.py
```diff
@@ -85,7 +87,9 @@ def apply_rough_alignment(render,
         # get the lowres stack rough alignment transformation
         tforms = lowres_ts[0].tforms
         tf = tforms[-1]
-        tf.M[0:2,0:2]*=scale
+        
+        if apply_scale:
+            tf.M[0:2,0:2]*=scale
```
RussTorres: I think this should have an else, doing something like 
```python
tf.M[:2, -1] /= scale
```
Which will account for the scaling of the solution's translation (unnecessary when scale is applied to the downsamples).

mahaling: The else would not be required if the scaling is not applied to downsamples... which will be the case from now on in the workflow... I am not sure I am following the "scaling of the solution's translation" here..

rendermodules/rough_align/apply_rough_alignment_to_montages.py
```diff
@@ -68,6 +69,7 @@ def apply_rough_alignment(render,
                           output_dir,
                           scale,
                           Z,
+                          apply_scale=False,
                           consolidateTransforms=True):
```
RussTorres: I don't think apply_scale is in the schema for this and it is never passed to this function.

PR #151, opened 2018-08-13
Extended qc plots

mahaling:
extended qc plots for visualizing point matches and montage residuals


rendermodules/em_montage_qc/plots.py
```diff
@@ -31,6 +117,11 @@ def plot_defects(render, stack, out_html_dir, args):
         pts.append([ts.minX, ts.maxY])
         pts.append([ts.minX, ts.minY])
         tile_positions.append(pts)
+        
+        if ts.tileId in tile_residual_mean.keys():
```
RussTorres: It's probably really expensive to have this iteration in a loop.  You can save a lot of time just catching a KeyError Exception or take it out of the loop with a set/view comparison operation.

PR #152, opened 2018-08-15
removing mipmap urls from downsample stack

mahaling:
1. removing mipmap urls from the downsample stack
2. adding a default value to "apply_scale" parameter in apply_rough_alignment_to_montages schema

### DIFF
--- a/rendermodules/dataimport/make_montage_scapes_stack.py
+++ b/rendermodules/dataimport/make_montage_scapes_stack.py
@@ -124,6 +124,7 @@ def create_montage_scape_tile_specs(render, input_stack, image_directory,
         t.width, t.height = im.size
     t.ip[0] = renderapi.image_pyramid.MipMap(
         imageUrl=pathlib.Path(filename).as_uri())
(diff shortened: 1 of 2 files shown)


rendermodules/dataimport/make_montage_scapes_stack.py
```diff
@@ -124,6 +124,7 @@ def create_montage_scape_tile_specs(render, input_stack, image_directory,
         t.width, t.height = im.size
     t.ip[0] = renderapi.image_pyramid.MipMap(
         imageUrl=pathlib.Path(filename).as_uri())
+    [t.ip.pop(k) for k in list(t.ip.keys()) if k != '0']
     t.minIntensity = 0
```
RussTorres: This fixes the bug, so I'm happy to merge it in.  What it doesn't do is address the improper use of the image pyramid which caused the issue in the first place.

For the sake of self-documenting code (and this being one of the few functions using the new imagepyramid interface) It would be better to explicitly assign a correct imagepyramid to t.ip via `renderapi.image_pyramid.ImagePyramid` or  `renderapi.image_pyramid.ImagePyramid.from_dict`.

PR #154, opened 2018-08-21
Parameterize mesh lc pointmatching

djkapner:


rendermodules/pointmatch/generate_point_matches_opencv.py
```diff
@@ -0,0 +1,286 @@
...
+    k2 = []
+    for m, n in matches:
+        if m.distance < args['ratio_of_dist']*n.distance:
+            good.append(m)
+    if len(good) > MIN_MATCH_COUNT:
+        src_pts = np.float32(
+                [k1xy[k1ind, :][m.queryIdx] for m in good]).reshape(-1, 1, 2)
+        dst_pts = np.float32(
+                [k2xy[m.trainIdx] for m in good]).reshape(-1, 1, 2)
+        M, mask = cv2.findHomography(
```
RussTorres: For the sake of utility, it could be nice to include a multiple hypothesis RANSAC option like Stephan has implemented in Java.  We can have that as a separate PR as well.

djkapner: I'm not familiar with that feature at the moment, so, I'd agree with separate PR.

PR #156, opened 2018-08-28
Mask corners lens corr

djkapner:
Allows masking in the mesh lens correction.
Remains to propagate mask through generate_mipmaps (separate PR)


rendermodules/mesh_lens_correction/do_mesh_lens_correction.py
```diff
@@ -159,8 +211,14 @@ def run(self):
...
 
+        self.maskUrl = make_mask(args_for_input)
+        # argshema doesn't like the NumpyArray after processing it once
+        # we don't need it after mask creation
+        self.args['mask_coords'] = None
+        args_for_input['mask_coords'] = None
+
         # create a stack with the lens correction tiles
-        ts_example = self.generate_ts_example()
+        ts_example = self.generate_ts_example(self.maskUrl)
```
RussTorres: is it necessary to have this as an argument if it is already an attribute of `self`?

djkapner:
thwarted by codecov!
This test, designed to get the last couple lines for read-only directories works for me locally, but not in Bamboo:
```
    # make the permissions of the mask_dir read-only
    os.chmod(mask_dir, 0o444)
    with pytest.raises(IOError):
        maskUrl3 = make_mask(
                mask_dir,
                width,
                height,
                mask_coords,
                mask_file=None,
                basename=None)
```
Suggestions welcome!

PR #157, opened 2018-08-29
Registration fusion

RussTorres:
implements simple bidirectional fusion defined by a json representation of chunks relations.  Currently implemented with interpolated render transformations and large transformLists for debugging -- it should be straightforward to use the concatenate render-module functionality to reduce the component transforms if that is easier.


rendermodules/fusion/fuse_stacks.py
```diff
@@ -0,0 +1,180 @@
...
+
+        resolved_tspecs = []
+        all_resolved_tform = []
+        for z in uninterpolated_zs:
+            section_tiles = []
+            resolved = renderapi.resolvedtiles.get_resolved_tiles_from_z(
+                child, z, render=self.render)
+            all_resolved_tform += resolved.transforms
+            for ts in resolved.tilespecs:
+                tsc = copy.deepcopy(ts)
```
fcollman: why is this necessary?

RussTorres: The copy operation is not strictly necessary -- I had been trying to keep the objects distinct for debugging purposes, but it shouldn't really change anything if we don't stop inside the iteration.

integration_tests/test_fusion.py
```diff
@@ -0,0 +1,288 @@
...
+        outd['transform'])
+    tspecs = renderapi.tilespec.get_tile_specs_from_stack(  # noqa: F841
+        childstack, render=render)
+    assert numpy.allclose(estimated_tform.M, tform.M)
+    assert outd['stack_a'] == parentstack
+    assert outd['stack_b'] == childstack
+
+
+def test_register_all_stacks(render, stack_DAG):
+    from rendermodules.fusion.register_adjacent_stack import (
```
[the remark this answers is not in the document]
RussTorres: @mahaling I think this was an effort to have module imports fail at the level of an individual test rather than the larger test module.  It doesn't match the pattern used in the rest of this repo, so I have changed the test.

PR #162, opened 2018-09-12
Lc masks into mipmap stacks

djkapner:
taking masks from lens correction, downsampling into mipmap levels, adding into stacks during apply_lens_correction()

### DIFF
--- a/rendermodules/lens_correction/apply_lens_correction.py
+++ b/rendermodules/lens_correction/apply_lens_correction.py
@@ -84,7 +84,7 @@ class ApplyLensCorrection(StackTransitionModule):
             self.input_stack, self.zValues[0], render=r)
         levels = [int(l) for l in tspecs[0].ip.levels]
         # make mask mipmaps
-        mask_mm_list = []
+        mask_mm_list = {}
         if self.args['maskUrl'] is not None:
             root, ext = os.path.splitext(self.args['maskUrl'])
             mask_mm_list = create_mipmaps(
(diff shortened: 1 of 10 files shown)


rendermodules/lens_correction/apply_lens_correction.py
```diff
@@ -75,6 +79,23 @@ def run(self):
...
             refId=lc_tform.transformId)
+
+        tspecs = renderapi.tilespec.get_tile_specs_from_z(
+            self.input_stack, self.zValues[0], render=r)
+        levels = [int(l) for l in tspecs[0].ip.levels]
+        # make mask mipmaps
+        mask_mm_list = []
+        if self.args['maskUrl'] is not None:
+            root, ext = os.path.splitext(self.args['maskUrl'])
+            mask_mm_list = create_mipmaps(
```
RussTorres: I think this defaults to using `skimage.measure.block_reduce` applying a `numpy.mean` function, which can lead to non-binary masked pixels.  A min, max, or median reduction function would probably work better for a mask.

rendermodules/lens_correction/apply_lens_correction.py
```diff
@@ -83,6 +104,8 @@ def run(self):
 
             for ts in tspecs:
                 ts.tforms = [ref_lc] + ts.tforms
+                for i in range(len(mask_mm_list)):
```
RussTorres: This makes an assumption about the levels available in the tilespec -- something like the following would fit the create_mipmaps file better
```python
for lvl, maskUrl in mask_mm_list.items():  # mask_mm_list is a dictionary
    ts.ip[lvl].maskUrl = pathlib.Path(maskUrl).as_uri()
```

djkapner:
rebased onto the passing master, should be good, barring other comments

PR #164, opened 2018-09-15
Swap zs

mahaling:
Swapping functionalities to swap zs in stacks and point match collections. 
Module needed for reimaging strategy


rendermodules/pointmatch/swap_point_match_collections.py
```diff
@@ -0,0 +1,97 @@
...
+        "host": "em-131db",
+        "port": 8080,
+        "owner": "TEM",
+        "project": "17797_1R",
+        "client_scripts": "/allen/aibs/pipeline/image_processing/volume_assembly/render-jars/production/scripts/"
+    },
+    "match_owner": "match_owner",
+    "source_collection": "source_collection",
+    "target_collection": "target_collection",
+    "zValues": []
```
fcollman: does an empty list mean do nothing? or swap them all? just some documentation, this example seems to imply to me subtly that an empty list will do them all, because why show an example that will do nothing.

rendermodules/stack/swap_zs.py
```diff
@@ -0,0 +1,104 @@
...
+
+example = {
+    "render": {
+        "host": "em-131db",
+        "port": 8080,
+        "owner": "TEM",
+        "project": "17797_1R",
+        "client_scripts": "/allen/aibs/pipeline/image_processing/volume_assembly/render-jars/production/scripts/"
+    },
+    "source_stack": "source_stack",
```
fcollman: this example is a single stack but the code has this a a list.

rendermodules/stack/swap_zs.py
```diff
@@ -0,0 +1,104 @@
...
+                                                                      host=self.render.DEFAULT_HOST,
+                                                                      port=self.render.DEFAULT_PORT,
+                                                                      render=self.render)
+        
+        source_stacks = []
+        target_stacks = []
+        zvalues = []
+        for source_stack, target_stack, zVals in zip(self.args['source_stack'], self.args['target_stack'], self.args['zValues']):
+            if source_stack not in list_of_stacks or target_stack not in list_of_stacks:
+                continue
```
fcollman: if the user provides a stack that isn't there don't we want to raise an exception rather than just completing doing nothing?

integration_tests/test_swapping.py
```diff
@@ -0,0 +1,176 @@
...
+        "target_stack": [target_stack],
+        "zValues": [[1015]],
+        "delete_source_stack": "False"
+    }
+
+    mod = swap_zs.SwapZsModule(input_data=ex, args=['--output_json', 'output.json'])
+    mod.run()
+
+    # assert that both the stack exist
+    list_of_stacks = renderapi.render.get_stacks_by_owner_project(host=render.DEFAULT_HOST,
```
fcollman: not sure why you included all the render.DEFAULT_HOST etc... i think that will just fill in by default.   This isn't wrong, just concerned there is something wrong with render-python that i'm not aware of.

integration_tests/test_swapping.py
```diff
@@ -0,0 +1,176 @@
...
+
+    assert(ts.tileId in target_tileids for ts in tspecs1)
+    assert(ts.tileId in source_tileids for ts in tspecs2)
+
+    ex['source_stack'] = [source_stack, target_stack]
+    ex['target_stack'] = [target_stack]
+
+    mod = swap_zs.SwapZsModule(input_data=ex, args=['--output_json', 'output.json'])
+    
+    with pytest.raises(RenderModuleException) as e:
```
fcollman: wasn't immediately obvious why this should fail.. i think because the target stack is both source and target.  can you add a comment?

rendermodules/stack/swap_zs.py
```diff
@@ -0,0 +1,104 @@
...
+            szvalues = renderapi.stack.get_z_values_for_stack(source_stack, render=self.render)
+            tzvalues = renderapi.stack.get_z_values_for_stack(target_stack, render=self.render)
+            avail_zs = set(szvalues).intersection(tzvalues)
+            final_zs = list(set(zVals).intersection(avail_zs))
+
+            source_stacks.append(source_stack)
+            target_stacks.append(target_stack)
+            zvalues.append(final_zs)
+
+        self.swap_zs_from_stacks(source_stack, target_stack, final_zs)
```
RussTorres: I think this is supposed to be inside the loop -- this will work for a single src and target, but not multiple.

.coveragerc
```diff
@@ -1,23 +0,0 @@
-[report]
```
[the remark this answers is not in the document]
mahaling: I did not... rebasing with master must have screwed up some files. will add it

integration_tests/test_swapping.py
mahaling: this usually throws an error for me from the ipython interface.. so adding it for extra caution... I do have the latest version of render-python...

rendermodules/stack/swap_zs.py
mahaling: these examples are dummy ones.. will add a real one...

rendermodules/stack/swap_zs.py
mahaling: Yes, but no here as this is a list of source and target stacks... we do not want to stop processing the list if we did not find one pair of source and target stacks. 
The output json will include the list of successfully processed source and target stacks for bookkeeping purposes. 

We could add a print statement if we want... But that's only for logging purposes

integration_tests/test_swapping.py
fcollman: okay.. thanks.. i'll take a look at render-python and see if i can figure this out, meanwhile no need to change this.

rendermodules/pointmatch/swap_point_match_collections.py
```diff
@@ -0,0 +1,97 @@
...
+    default_schema = SwapPointMatches
+    default_output_schema = SwapPointMatchesOutput
+
+    def run(self):
+        # get match collections 
+        collecs = renderapi.pointmatch.get_matchcollections(owner=self.args['match_owner'], render=self.render)
+
+        collections = [c['collectionId']['name'] for c in collecs]
+
+        print(self.args['source_collection'], self.args['target_collection'])
```
fcollman: these should be logger statements I think.

rendermodules/stack/swap_zs.py
fcollman: so if both source and target are empty should we raise an exception then?

fcollman: err.. sorry.. after the for loop if the source_stacks and target_stacks are still empty.

rendermodules/mesh_lens_correction/schemas.py
```diff
@@ -73,10 +73,25 @@ class MeshLensCorrectionSchema(PointMatchOpenCVParameters):
     z_index = Int(
         required=True,
         description="z value for the lens correction data in stack")
+    nfeature_limit = Int(
```
RussTorres: These should probably be removed as they are now defined in the opencv pointmatch module.

PR #165, opened 2018-09-21
Meshlc py tps and affine

djkapner:
this PR has 2 components:
1) replaces java client ThinPlateSplineClient call with new render-python renderapi.transform.ThinPlateSplineTransform.estimate() call.
2) in MeshAndSolveTransform, a common affine is collected from the solve results and included in the resulting transform. This will allow stacks using these transforms to be solved without affine (e.g. rigid, similarity, translation...)


fcollman:
This looks good to me

PR #173, opened 2018-10-11
Match filtering

djkapner:
I am finding this very useful for looking at point match results.
I think this should be part of QC in the workflow, somehow.
In the meantime, this module is 100% tested. I thought I'd PR it to get it some circulation.

figure_1-9


djkapner:
What it does:
- takes the point match results and fits an AffineModel to the matches (pairwise, this is not a global solve)
- looks at residuals per tile pair, and, how far the fitted translation is from the input_stack (stage coordinates, generally)
- outputs some summary
- optionally, can write 0-weights back to the collection, or a collection copy, which the solver will later ignore.

PR #176, opened 2018-10-26
bumping EMaligner, adding test for finer control of polynomial regula…

djkapner:
…rization

EMaligner now has finer tuned control for polynomial parameter regularization.
For example, for 2nd order polynomial, a regularization.poly_factors list of 
`[1e-5, 1.0, 1e6]` specifies factors for 0th, 1st, and 2nd order parameters respectively.
Each of these are multiplied by regularization.default.lambda, keeping with tradition of having default_lambda and translation_factor for affine, which is equivalent to a 1st order polynomial with poly_factors `[translation_factor, default_lambda]`

The purpose is to add some higher constraint to the 2nd order parameters, as, for poorly matched point matches, these parameters can get large enough to drastically distort a tile, usually resin or blood vessel. Specifying something like 1e6 prevents these higher order corrections from becoming too powerful.


djkapner:
rebased to master

PR #177, opened 2018-10-30
Overwrite montage scape section

mahaling:
Functionality to overwrite existing Z in montage scape stack. This avoids multiple tiles in same Z issue on re-runs.

### DIFF
--- a/rendermodules/dataimport/make_montage_scapes_stack.py
+++ b/rendermodules/dataimport/make_montage_scapes_stack.py
@@ -214,6 +214,32 @@ class MakeMontageScapeSectionStack(StackOutputModule):
 
         Z = [[int(oldz),int(newz)] for oldz, newz in zip(zvalues, newzvalues)]
 
+        out_stack_exists = self.args['output_stack'] in self.render.run(renderapi.render.get_stacks_by_owner_project)
+
+        if out_stack_exists:
+            # check whether overwrite z is set to false. If so, then remove those z that is already in output stack
+            outzvalues = renderapi.stack.get_z_values_for_stack(
+                            self.args['output_stack'],
+                            render=self.render)
+            
+            if self.args['overwrite_z'] is False:
+                # do not include those sections that are already in the output stack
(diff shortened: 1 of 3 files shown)


rendermodules/dataimport/make_montage_scapes_stack.py
```diff
@@ -214,6 +214,32 @@ def run(self):
...
+
+        if out_stack_exists:
+            # check whether overwrite z is set to false. If so, then remove those z that is already in output stack
+            outzvalues = renderapi.stack.get_z_values_for_stack(
+                            self.args['output_stack'],
+                            render=self.render)
+            
+            if self.args['overwrite_z'] is False:
+                # do not include those sections that are already in the output stack
+                Z = [[int(oldz), int(newz)] for oldz, newz in zip(zvalues, newzvalues) if not(float(newz) in outzvalues)]
```
RussTorres: Is this the behavior we want?  The other modules treat overwrite_zlayer as optionally deleting what is there vs keeping it and overlaying (I think the idea was for multiple montage sets mapping to a single section).  There  are situations where this way would probably help more, but it is a departure.

mahaling: I agree... this comes useful when handling torn sections.. Will make the changes and commit.

RussTorres:
@gayathrimahalingam It's good to look at this code again -- it seems that this code triggers state transitions from LOADING to COMPLETE and does not obey the close_stack parameter (because it doesn't use the StackOutputModule's tilespec-based import).  I think I've seen workflow failures because of this, and since `self.close_stack` is already a parameter it should be easy to include.

mahaling:
> @gayathrimahalingam It's good to look at this code again -- it seems that this code triggers state transitions from LOADING to COMPLETE and does not obey the close_stack parameter (because it doesn't use the StackOutputModule's tilespec-based import). I think I've seen workflow failures because of this, and since `self.close_stack` is already a parameter it should be easy to include.

I am surprised that we still got this code working so far in the workflow without encapsulation of close_stack parameter check with setting the state of the stack. Encapsulated it now.

PR #181, opened 2018-11-09
Error checks on swaps

mahaling:
Added some error checks wherever possible.. Still not 100% fail proof due to render client api call fail scenario.


rendermodules/pointmatch/swap_point_match_collections.py
```diff
@@ -39,51 +116,22 @@ def run(self):
...
-                                                    sourcematches,
-                                                    owner=self.args['match_owner'],
-                                                    render=self.render)
-                # import target matches into source collection
-                renderapi.pointmatch.import_matches(self.args['source_collection'],
-                                                    targetmatches,
-                                                    owner=self.args['match_owner'],
-                                                    render=self.render)
-                zvalues.append(z)
+        zvalues = [z for z, n in zip(ids, output_bool) if output_bool ]
```
RussTorres: This should test for n:
```suggestion
        zvalues = [z for z, n in zip(ids, output_bool) if n]    
```

rendermodules/pointmatch/swap_point_match_collections.py
```diff
@@ -17,6 +19,81 @@
     "zValues": [1015]
 }
 
+def swap_pt_matches(render, source_collection, target_collection, z, match_owner=None):
+    session = requests.Session()
+    session.mount('http://', requests.adapters.HTTPAdapter(max_retries=5))
+
+    groupid = "{}.0".format(str(z))
+    match_owner = match_owner if match_owner is not None else render.DEFAULT_OWNER
```
RussTorres: This doesn't need to change, but the logic should be handled by render-python -- owner in a call overrides the render default owner.

mahaling: yep this just adds a double check

rendermodules/stack/swap_zs.py
```diff
@@ -17,52 +19,99 @@
...
+                        source_stack,
+                        "LOADING",
+                        session=session)
+        render.run(renderapi.client.import_tilespecs_parallel,
+                        source_stack,
+                        target_ts.tilespecs,
+                        sharedTransforms=target_ts.transforms)
+    except Exception as e:
+        print("Cannot import tilespecs to {} from {}".format(source_stack, target_stack))
+        return temp_stacks, False
```
RussTorres: This will evaluate as `True` in line 152.

PR #182, opened 2018-11-15
Safe swapping

mahaling:
Changing the swap module to first copy and then delete, rather than delete and then copy.

### DIFF
--- a/rendermodules/stack/swap_zs.py
+++ b/rendermodules/stack/swap_zs.py
@@ -105,8 +105,8 @@ class SwapZsModule(RenderModule):
             avail_zs = set(szvalues).intersection(tzvalues)
             final_zs = list(set(zVals).intersection(avail_zs))
 
-            source_stacks.append(source_stack)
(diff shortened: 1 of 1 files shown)


rendermodules/stack/swap_zs.py
```diff
@@ -34,110 +34,51 @@ def swap_section(render, source_stack, target_stack, z, pool_size=5):
...
+    target_tileids = [ts.tileId for ts in target_ts.tilespecs]
 
-    try:
-        render.run(renderapi.stack.create_stack,
-                   temp_stack1)
-        render.run(renderapi.stack.create_stack,
-                   temp_stack2)
-    except Exception as e:
-        print("Cannot create temp stacks")
+    if set(source_tileids) == set(target_tileids):
```
RussTorres: This might even need a stronger assertion, since any duplicate tileids will be lost in both stacks
```suggestion
    if set(source_tileids) & set(target_tileids):
```

PR #188, opened 2019-03-13
Downsample masks

djkapner:
* One of the steps in improving rough alignment (Jan/Feb 2019) was to mask off some split sections and also ignore the point matches in the masked areas.
* Previously, we had already added mask handling in apply_rough. This PR is just giving some functionality for adding masks and filtering point matches. 
* The intent is for it to run out-of-workflow (open to debate...). I found that I had to experiment with what masks to include and which ones were not necessary. So, I ended up having something that put them in and took them out, both for the stack and the collection. This is just a clean version of that manual/scripty thing I was doing.
* Some time ago, we had the experience that RANSAC automagically chose one half of a split section. That turned out not to hold true always, so I think the point match filter part of this is necessary for clean rough alignment.
* Also, pytest is up to 4.0 and that broke an unrelated test in test_lens_correction, so that explains the last commit and why master build is currently broken.

### DIFF
--- a/integration_tests/test_downsample_mask.py
+++ b/integration_tests/test_downsample_mask.py
@@ -87,10 +87,7 @@ def matches_are_filtered(render, collection):
                 collection, group1, group2, render=render)
     ids = []
     for m in matches:
-        if np.count_nonzero(
-                np.isclose(
-                    np.array(m['matches']['w']),
-                    0.0)) != 0:
+        if np.any(np.isclose(np.array(m['matches']['w']), 0.0)):
             ids.append([m['pGroupId'], m['qGroupId']])
     return ids
 
--- a/integration_tests/test_downsample_mask.py
+++ b/integration_tests/test_downsample_mask.py
@@ -6,11 +6,12 @@ from test_data import (
         render_params)
 
 from rendermodules.rough_align.downsample_mask_handler import (
-        DownsampleMaskHandler, points_in_mask)
+        DownsampleMaskHandler, points_in_mask, polygon_list_from_mask)
 from rendermodules.module.render_module import \
         RenderModuleException
 import pytest
 import numpy as np
+from shapely.geometry import Polygon, Point
 
 DS_MASK_TEST_PATH = os.path.join(
         TEST_DATA_ROOT,
@@ -94,6 +95,51 @@ def matches_are_filtered(render, collection):
     return ids
 
 
+def test_mask_polygons():
+    mask1 = np.zeros((100, 100)).astype('uint8')
+    mask2 = np.zeros((100, 100)).astype('uint8')
+    mask1[40:60, 40:60] = 255
+    mask2[0:10, 20:50] = 255
+    polys = polygon_list_from_mask(mask1 + mask2)
+    assert len(polys) == 2
+
+    # flip the axes order to be (col, row), i.e. (x, y)
+    m1mnmx = np.hstack((
+        np.argwhere(mask1 == 255).min(axis=0)[::-1],
+        np.argwhere(mask1 == 255).max(axis=0)[::-1]))
+    m2mnmx = np.hstack((
+        np.argwhere(mask2 == 255).min(axis=0)[::-1],
+        np.argwhere(mask2 == 255).max(axis=0)[::-1]))
+
+    def check_matching(polys, m1mnmx, m2mnnx):
+        # check that the polygons match up 1-to-1 with the input
+        match = [-1] * len(polys)
+        for i in range(len(polys)):
+            pmnmx = np.hstack((
+                np.array(polys[i].boundary.xy).min(axis=1).astype('int'),
+                np.array(polys[i].boundary.xy).max(axis=1).astype('int')))
+            if np.all(np.isclose(pmnmx, m1mnmx)):
+                match[i] = 1
+            if np.all(np.isclose(pmnmx, m2mnmx)):
+                match[i] = 2
+        assert (match == [1, 2]) | (match == [2, 1])
+
+    check_matching(polys, m1mnmx, m2mnmx)
+
+    # give a transform
+    B0 = 10
+    B1 = -20
+    tf = renderapi.transform.AffineModel(
+            B0=B0,
+            B1=B1)
+    polys = polygon_list_from_mask(mask1 + mask2, transforms=[tf])
+    m1mnmx[[0, 2]] += B0
+    m1mnmx[[1, 3]] += B1
+    m2mnmx[[0, 2]] += B0
+    m2mnmx[[1, 3]] += B1
+    check_matching(polys, m1mnmx, m2mnmx)
+
+
 def test_points_in_mask():
     mask = np.zeros((1000, 1000)).astype('uint8')
     # render masks areas where mask = 0
@@ -109,8 +155,8 @@ def test_points_in_mask():
     # all should be weight=0 (excluded)
     w0pts = np.random.rand(2, 100)
     w0pts[0, :] *= 1000
-    w0pts[1, :] *= 500
-    w0pts[1, :] += 500
+    w0pts[1, :] *= 400
+    w0pts[1, :] += 510
     w0 = points_in_mask(mask, w0pts.tolist())
     assert np.all(np.isclose(w0, 0.0))
 
--- a/integration_tests/test_downsample_mask.py
+++ b/integration_tests/test_downsample_mask.py
@@ -1,0 +1,219 @@
+import os
+import renderapi
+import json
+from test_data import (
+        TEST_DATA_ROOT,
+        render_params)
+
+from rendermodules.rough_align.downsample_mask_handler import (
+        DownsampleMaskHandler, points_in_mask)
+from rendermodules.module.render_module import \
+        RenderModuleException
+import pytest
+import numpy as np
+
+DS_MASK_TEST_PATH = os.path.join(
+        TEST_DATA_ROOT,
+        "em_modules_test_data",
+        "downsample_masking")
+DS_MASK_PATH = os.path.join(
+        DS_MASK_TEST_PATH,
+        "masks")
+DS_BAD_MASK_PATH = os.path.join(
+        DS_MASK_TEST_PATH,
+        "bad_masks")
+DS_TSPEC_PATH = os.path.join(
+        DS_MASK_TEST_PATH,
+        "ds_tilespecs.json")
+DS_COLLECTION_PATH = os.path.join(
+        DS_MASK_TEST_PATH,
+        "ds_matches.json")
+
+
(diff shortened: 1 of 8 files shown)


djkapner:
the commit history looks a bit odd here...
I based off master, merged one change from develop (the pytest thing), but the number of commits and files is off...
did somehow master get ahead of develop?

integration_tests/test_downsample_mask.py
```diff
@@ -0,0 +1,219 @@
...
+    return zs
+
+
+def matches_are_filtered(render, collection):
+    groups = renderapi.pointmatch.get_match_groupIds(
+            collection,
+            render=render)
+    matches = []
+    for i in range(len(groups)):
+        for j in range(i):
```
RussTorres: This is probably more readable as `for group1, group2 in itertools.combinations(groups, 2):`, which is ordered differently but shouldn't matter in the renderapi call.

rendermodules/rough_align/downsample_mask_handler.py
```diff
@@ -0,0 +1,192 @@
...
+        tspec = self.get_tilespec(z)
+        if not tspec:
+            return tspec
+
+        if tspec.ip[0].maskUrl is not None:
+            raise RenderModuleException(
+                    "Tilespec already has a mask for z = %d" % (z))
+
+        fnames = glob.glob(os.path.join(
+            self.args['mask_dir'], '%d_*.png' % z))
```
RussTorres: Does this module only accept "png" masks while the other one takes "png" or "tif"?

integration_tests/test_downsample_mask.py
```diff
@@ -0,0 +1,219 @@
...
+            matches += renderapi.pointmatch.get_matches_from_group_to_group(
+                    collection, groups[i], groups[j], render=render)
+    ids = []
+    for m in matches:
+        if np.count_nonzero(
+                np.isclose(
+                    np.array(m['matches']['w']),
+                    0.0)) != 0:
```
mahaling: np.isclose returns a boolean... is there a specific reason that you are checking for "!=0" rather than "==1"?

mahaling: ok !=0 is for the count_nonzero... got it.

mahaling: could change that to >0

integration_tests/test_rough_align.py
```diff
@@ -417,8 +422,31 @@ def test_do_rough_alignment_with_scale(render, montage_scape_stack_with_scale, r
...
+@pytest.fixture(scope='module')
+def test_do_rough_alignment_python(
+        render, montage_scape_stack, rough_point_match_collection,
+        tmpdir_factory, output_lowres_stack=None):
+
+    if output_lowres_stack == None:
+        output_lowres_stack = '{}_DS_Rough_python'.format(montage_scape_stack)
+    output_directory = str(tmpdir_factory.mktemp('output_json'))
+
+    solver_input = copy.deepcopy(test_rough_parameters_python)
```
mahaling: why do you need a deepcopy here?

rendermodules/rough_align/apply_rough_alignment_to_montages.py
```diff
@@ -61,39 +66,157 @@
...
+        for t in tilespecs:
+            if t.ip[0].maskUrl is not None:
+                results[t.tileId] = t.ip[0].maskUrl
+
+    elif mask_input_dir is not None:
+        tids = [t.tileId for t in tilespecs]
+        maskfiles = []
+        for ext in exts:
+            e = mask_input_dir + '/*.' + ext
+            maskfiles += glob.glob(e)
```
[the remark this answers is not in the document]
mahaling: I agree... We cannot assume that masks will be stored in the same directory. We can either document this assumption here and revisit later to fix it.

rendermodules/rough_align/apply_rough_alignment_to_montages.py
```diff
@@ -61,39 +66,157 @@
...
+        maskbasenames = np.array([os.path.basename(os.path.splitext(p)[0])
+                                  for p in maskfiles])
+        for t in tids:
+            ind = np.argwhere(maskbasenames == t).flatten()
```
mahaling: Is this an assumption that mask names represent the tileids' name?.

rendermodules/rough_align/apply_rough_alignment_to_montages.py
```diff
@@ -61,39 +66,157 @@
...
+        for tf in tspec_lowres.tforms:
+            c = tf.tform(c)
+        mask_polygons.append(Polygon(c).buffer(0))
+
+    new_highres = []
+    for t in resolved_highres.tilespecs:
+        tc = t.bbox_transformed(
+                    reference_tforms=resolved_highres.transforms)
+        tpoly = Polygon(tc).buffer(0)
+        for p in mask_polygons:
```
mahaling: this could be a time consuming for loop.. Have you thought about KDtree or quadtree to find the intersection of a set of polygons at a time.

rendermodules/rough_align/apply_rough_alignment_to_montages.py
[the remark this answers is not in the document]
djkapner: - just FYI, the file `apply_rough_alignment_to_montages.py` is not part of this PR. It is showing up because master got ahead of develop (my bad).
- nonetheless, good point, I'll change this. Point of reference, for the 20k sections, we had about fewer than 50 masks... not sure how much we need to worry about performance issues here.
- @gayathri, it is implicit from the schema that there is one mask directory. If there are multiple directories, we can run the module multiple times. Also, it is likely we won't be reading from directory at this point as we had added `read_masks_from_lowres_stack` as an option and the whole PR here is to get the masks into the lowres stack so we can use that.

rendermodules/rough_align/downsample_mask_handler.py
```diff
@@ -0,0 +1,192 @@
...
+                    render=self.render)
+
+        # remove any masking for each zReset
+        tspecs = []
+        for z in np.setdiff1d(self.args['zReset'], None):
+            tspec = self.get_tilespec(z)
+            if tspec:
+                tspecs.append(tspec)
+                if tspec:
+                    tspecs.append(tspec)
```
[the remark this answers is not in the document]
djkapner: I wish I could say this is an elegant, little known pythonic contsruction that does something really cool...
It was a typo

rendermodules/rough_align/apply_rough_alignment_to_montages.py
```diff
@@ -61,39 +66,157 @@
...
+
+    if tspec_lowres.tileId not in mask_map.keys():
+        return resolved_highres.tilespecs
+
+    impath = urllib.parse.unquote(
+                 urllib.parse.urlparse(
+                     mask_map[tspec_lowres.tileId]).path)
+    maskim = cv2.imread(impath, 0)
+    maskim = 255 - maskim
+    _, contours, _ = cv2.findContours(
```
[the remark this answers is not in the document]
djkapner: I cleaned it up by sharing some code between the 2.
In answer to your question, I think yes, except for some edge cases and rounding issues. The DS masks do line up exactly with some of the high res tiles on some edges, so, this way of `not intersect with the masked part` for including was a good way to catch all that. I added some comments to that affect, and, there are new doc strings for the underlying mask-related functions.

rendermodules/rough_align/schemas.py
```diff
@@ -71,18 +77,46 @@ class ApplyRoughAlignmentTransformParameters(RenderParameters):
         Int,
         required=True,
         cli_as_single_argument=True,
-        description="List of z values to apply rough alignment to")  
-    
+        description="List of z values to apply rough alignment to")
+    mask_input_dir = InputDir(
```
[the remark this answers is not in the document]
djkapner: as doc strings now in `polygon_list_from_mask` and `points_in_mask`

integration_tests/test_downsample_mask.py
djkapner: agreed

rendermodules/rough_align/downsample_mask_handler.py
djkapner: mirrored that same `mask_exts` schema entry

integration_tests/test_downsample_mask.py
djkapner: changed to use np.any() to be more readable

integration_tests/test_rough_align.py
djkapner: our parameters tend to be nested dictionaries. Things get messy with a shallow copy:
```
In [247]: import json

In [248]: print(json.dumps(a, indent=2))
{
  "b": {
    "c": 1
  }
}

In [249]: aa = dict(a)

In [250]: print(json.dumps(aa, indent=2))
{
  "b": {
    "c": 1
  }
}

In [251]: a['b']['c'] = np.pi

In [252]: print(json.dumps(aa, indent=2))
{
  "b": {
    "c": 3.141592653589793
  }
}

In [253]: xx = copy.deepcopy(a)

In [254]: print(json.dumps(xx, indent=2))
{
  "b": {
    "c": 3.141592653589793
  }
}

In [255]: a['b']['c'] = 1

In [256]: print(json.dumps(xx, indent=2))
{
  "b": {
    "c": 3.141592653589793
  }
}

In [257]: print(json.dumps(a, indent=2))
{
  "b": {
    "c": 1
  }
}
```

rendermodules/rough_align/apply_rough_alignment_to_montages.py
djkapner: already in apply rough schema:
```
    mask_input_dir = InputDir(
        required=False,
        default=None,
        missing=None,
        description=("directory containing mask files. basenames of "
                     "masks that match tileIds in the rough stack "
                     "will be handled."))
```

rendermodules/rough_align/apply_rough_alignment_to_montages.py
djkapner: It takes about 4 seconds per section with a mask.We had 28 masks in 20,000 sections. I don't think it's worth getting fancy here

integration_tests/test_rough_align.py
```diff
@@ -417,8 +422,31 @@ def test_do_rough_alignment_with_scale(render, montage_scape_stack_with_scale, r
     assert(set(zvalues) == set(zs))
 
     yield output_lowres_stack
-    renderapi.stack.delete_stack(output_lowres_stack, render=render)
+    #renderapi.stack.delete_stack(output_lowres_stack, render=render)
```
RussTorres: If things are mostly looking good can we un-comment these lines?

PR #189, opened 2019-03-15
remap zs change in rough alignment workflow

mahaling:
Changing the functionality of automatically remapping section ids to the z value. 
1. Make montage scapes stack has a new parameter "remap_section_ids" a boolean which is by default "False" (an optional parameter)
2. Apply rough alignment has a new parameter "remap_section_ids" - again by default "False" and an optional parameter.
3. Changes have been made to rough align test to deprecate matlab solver based testing and rough alignment.


integration_tests/test_rough_align.py
```diff
@@ -534,192 +464,40 @@ def test_apply_rough_alignment_transform(render, montage_stack, test_do_rough_al
...
-
-    tspecs = render.run(
-        renderapi.tilespec.get_tile_specs_from_stack, output_stack)
-
-    tsfn = urllib.parse.unquote(urllib.parse.urlparse(
-        tspecs[0].ip[0].imageUrl).path)
-    assert os.path.isfile(tsfn)
-    assert os.path.basename(tsfn) == '1020.0.png'
-
-def test_apply_rough_alignment_transform_with_scale(render, montage_stack, test_do_rough_alignment_with_scale, tmpdir_factory, prealigned_stack=None, output_stack=None):
```
RussTorres: Why is this test removed?  If we don't want backwards-compatible scaling we should remove it from the module in another PR.

integration_tests/test_rough_align.py
```diff
@@ -534,192 +464,40 @@ def test_apply_rough_alignment_transform(render, montage_stack, test_do_rough_al
...
-        renderapi.stack.set_stack_state(ex2['lowres_stack'],'LOADING',render=render)
-        renderapi.stack.delete_section(ex2['lowres_stack'],1021,render=render)
-        renderapi.stack.set_stack_state(ex2['lowres_stack'],'COMPLETE',render=render)
-        mod = ApplyRoughAlignmentTransform(input_data=ex2, args=[])
-        mod.run()
-        zvalues = renderapi.stack.get_z_values_for_stack(ex2['output_stack'],render=render)
-        assert(1021 not in zvalues)
-
-
-def make_montage_stack_without_downsamples(render, montage_stack, tmpdir_factory):
```
RussTorres: This functionality should be dropped from the modules before it is left untested.  The complexity of the codebase is such that code coverage can be okay even if major things are untested.

integration_tests/test_rough_align.py
```diff
@@ -322,179 +323,108 @@ def rough_mapped_pt_match_collection(render, rough_mapped_pt_matches_from_json):
...
-        'source_collection': dict(solver_example['source_collection'], **{
-            'stack': montage_z_mapped_stack}),
-        'target_collection': dict(solver_example['target_collection'], **{
-            'stack': output_lowres_stack}),
-        'source_point_match_collection': dict(
-            solver_example['source_point_match_collection'], **{
-                'match_collection': rough_mapped_pt_match_collection
-            })
-    })
+    solver_input['first_section'] = 251
```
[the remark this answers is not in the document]
mahaling: ok will change that to create a copy of the original input.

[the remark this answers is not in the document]
djkapner: nested dicts need deepcopy

integration_tests/test_rough_align.py
mahaling: I removed it without making it backwards-compatible... The python solver removes the scale regardless of introducing it in the downsample stack (@djkapner please correct me if I am wrong). 
Will re-introduce the matlab solving options then.

djkapner: the python solver doesn't actively do anything with scale.
I think what you're talking about is that the matlab solver did something like scale the transforms for downsampled images up, which is nice if you wanted to ndviz downsampled sections on top of high res sections.
I don't do that in the python one. If we wanted that, I don't think that is a solver operation.

rendermodules/rough_align/apply_rough_alignment_to_montages.py
```diff
@@ -70,10 +71,11 @@ def apply_rough_alignment(render,
                           scale,
                           Z,
                           apply_scale=False,
-                          consolidateTransforms=True):
+                          consolidateTransforms=True,
+                          remap_section_ids=False):
     z = Z[0] # z value from the montage stack - to be mapped to the newz values in lowres stack
     newz = Z[1] # z value in the lowres stack for this montage
-    
+    print(z, newz)
```
djkapner: substitute logger.info or debug for print() so we can see this in the workflow logs?

integration_tests/test_rough_align.py
mahaling: The scale factor was added for the AT team's operations as they wanted to see the ndviz downsampled section on top of montages. We have never used such a functionality. The one scenario where this test would fail is when we apply scale in downsampled stack and then the python solver ignores it and we end up applying transformations with scale in apply rough... 
The module needs  a change when we completely switch to python solver (again to be done in another PR)

rendermodules/rough_align/apply_rough_alignment_to_montages.py
mahaling: removed it. This message will be a small fish out in the sea of logs... Not worth adding.

integration_tests/test_rough_align.py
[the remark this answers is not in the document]
mahaling: I am surprised nobody noticed this before... This has been the way for so long... Changing it to deepcopy.

integration_tests/test_rough_align.py
mahaling: fucntionality added again

integration_tests/test_rough_align.py
```diff
@@ -772,72 +956,6 @@ def test_setting_new_z_montage_scape(render, montage_stack, downsample_sections_
     assert(set(zvalues) == set([1020, 1021, 1022]))
 
 
-
-def test_solver_default_options(render, montage_scape_stack, rough_point_match_collection, tmpdir_factory):
```
RussTorres: There are reasons to keep some compatibility with the MATLAB solver, even if it is not used in production (and even more reasons not to rewrite tests/dig through PRs).  For things like this, could we mark the test with a skip rather than delete it for the immediate future and follow up with a conversation about how we want to deprecate things?

rendermodules/rough_align/schemas.py
```diff
@@ -85,6 +90,9 @@ def validate_data(self, data):
                 raise ValidationError("new_z list count does not match with old_z list count")
         else:
             data['new_z'] = data['old_z']
+        
+        if data['map_z'] is False:
```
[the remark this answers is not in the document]
mahaling: changed this.

integration_tests/test_rough_align.py
mahaling: added the test back

PR #191, opened 2019-03-20
Pairwise anchor rough

djkapner:
2nd of 2 updates to rough alignment. This one is ideally merged after downsample_masks.
2 main components here:
- Our original strategy for rough alignment was a global (EMaligner) similarity solve followed by an affine solve. The similarity solve needed to be loose enough (in regularization) to accommodate rotated montages coming out of imaging, but that looseness also allowed tile shrinkage of 10% or more while rotating the tiles.
- The first try to correct this was to just use a moving pairwise rigid alignment through the volume and follow that with a global affine solve. This generally worked to get rid of the scale changes, but with such a large volume, there was some accumulated unphysical drift.
- To correct the drift issue, Nuno made by-hand an alignment using trakem every 50 sections. These aligned sections were used as anchor points for the pairwise rigid alignment.
- (when EMaligner is working with true rigid transforms, this whole step might not be needed, but that is TBD.)

Contents:
- `render-modules/rendermodules/rough_align/make_anchor_stack.py ` takes transforms from the manual alignment and creates an anchor stack. This is meant to be out-of-workflow. It doesn't happen more than once, I think. This is an optional intermediate stack used by:
- `render-modules/rendermodules/rough_align/pairwise_rigid_rough.py ` moving pairwise rigid alignment through a downsampled stack. Tiles in between anchor points are appropriately weighted between the two.


rendermodules/rough_align/pairwise_rigid_rough.py
```diff
@@ -0,0 +1,282 @@
...
+                axis=0) for t in tilespecs]).min(axis=0)
+    for t in tilespecs:
+        t.tforms[0].M[0, 2] += \
+            translation_buffer[0] - xymin[0]
+        t.tforms[0].M[1, 2] += \
+            translation_buffer[1] - xymin[1]
+    return tilespecs
+
+
+class PairwiseRigidRoughAlignment(StackInputModule, StackOutputModule):
```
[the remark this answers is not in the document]
djkapner: ok, changed that. though I'm not seeing that I was using any custom "stack" parameters for this one...

rendermodules/rough_align/schemas.py
```diff
@@ -1,12 +1,113 @@
...
+
+
+class MakeAnchorStackSchema(StackTransitionParameters):
+    transform_json = InputFile(
+        required=True,
+        description=("Human generated list of transforms."
+                     "Keys are of form <z>_*.png where z matches "
+                     "a tilespec in input_stack and values are "
+                     "AffineModel transform jsons"))
+    zValues = List(
```
RussTorres: This should already be in the StackTransitionParameters (automatically generated based on `minZ` and `maxZ` if only inputting those)

djkapner:
> Is there also something to check in w.r.t. generating transformation files for the anchor stack? Are the transformations coming from a trakem2 xml? In which case it might be helpful for the future to have a module that can either generate input for MakeAnchorStack or invoke its run method based on xml input.

I forgot about that step... was kind of surprised how clean the input json was for this. Yes, I have these few functions from you I can work in:
```
import sys

from lxml import etree
import renderapi


def get_affinemodel_from_tformstr(s):
    return renderapi.transform.AffineModel(
        json={"dataString": s.lstrip("matrix(").rstrip(")").replace(",", " "),
              "className": 'mpicbg.trakem2.transform.AffineModel2D'})


def get_patch_transforms_from_tree(t):
    return {p.attrib['title']: get_affinemodel_from_tformstr(
        p.attrib["transform"]) for p in t.findall('.//t2_patch')}


def get_patch_transforms_from_xml(x):
    return get_patch_transforms_from_tree(etree.parse(x))


if __name__ == "__main__":
    xmlfn = sys.argv[1]
    outfn = sys.argv[2]

    t_to_tform_map = get_patch_transforms_from_xml(xmlfn)
    with open(outfn, 'w') as f:
        renderapi.utils.renderdump(t_to_tform_map, f, indent=2)
```

djkapner:
> I would prefer to have the pairwise rigid tests to be in a separate test file.. test_rough_align is getting crowded for more than what it needs

done

rendermodules/rough_align/pairwise_rigid_rough.py
```diff
@@ -0,0 +1,282 @@
...
+            result = {}
+            result['residuals'] = []
+            for r in pool.map(check_func, fargs):
+                    result['residuals'].append(r['avg_residual'])
+
+        zall = np.arange(
+                self.args['minZ'],
+                self.args['maxZ'] + 1)
+        z = np.array([t.z for t in tilespecs])
+        result['missing'] = np.setdiff1d(zall, z).tolist()
```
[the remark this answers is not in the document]
djkapner: added more words to the schema about this.
yes
this was the part of alignment where I was skipping sections if I could not find a way to align them, so the gap file is a way to introduce a skipped section.
I realize that it isn't necessarily synced up with the workflow "gap" sections, but, I needed a way to iterate over possibilities to look for alignment improvements...

rendermodules/rough_align/schemas.py
djkapner: I wanted to use the stack transition module, but, there's no meaning to minZ and maxZ relative to the input stack in this case, so I needed something here to keep the parent class happy.

RussTorres: I see, I didn't think about the use case with an external specification of zs.  Going to resolve this, but it's likely a quick modification of the `ZValueParameters` to support this.

rendermodules/rough_align/make_anchor_stack.py
```diff
@@ -0,0 +1,83 @@
+import renderapi
+import json
+from rendermodules.module.render_module import (
+        StackTransitionModule, RenderModuleException)
+from rendermodules.rough_align.schemas import (
+        MakeAnchorStackSchema)
+from lxml import etree
```
RussTorres: I don't think lxml is in the requirements file, but since it's technically not necessary you could also do something like this and call it out as an optional dependency.
```python
try:
    from lxml import etree
except ImportError:
    can_xml = False
```

And then later do 
```python
# when you need to parse the xml
if not can_xml:
    raise RenderModuleException("xml parsing requires lxml")
```

I also don't have anything against making it a hard requirement.

djkapner: I'd rather do hard requirement.

PR #196, opened 2019-03-29
adding validation to apply mipmaps

mahaling:
Apply mipmaps was not ingesting some tiles in some sections during Phase 2, which was detected at the montage manual qc stage. 

Adding the check in apply mipmaps, that produces the list of Zs for which apply mipmaps failed.. 
This can be consumed in the workflow strategy to fail the job for each montage based on this output.

### DIFF
--- a/rendermodules/dataimport/apply_mipmaps_to_render.py
+++ b/rendermodules/dataimport/apply_mipmaps_to_render.py
@@ -113,7 +113,7 @@ class AddMipMapsToStack(StackTransitionModule):
         for z in zvalues:
             input_rs = renderapi.resolvedtiles.get_resolved_tiles_from_z(input_stack, z, render=self.render)
             output_rs = renderapi.resolvedtiles.get_resolved_tiles_from_z(output_stack, z, render=self.render)
-            if len(input_rs.tilespecs) != len(output_rs.tilespecs):
+            if len(input_rs.tilespecs) != len(output_rs.tilespecs): # pragma: no cover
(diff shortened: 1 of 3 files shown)


rendermodules/dataimport/apply_mipmaps_to_render.py
```diff
@@ -102,7 +102,22 @@ def run(self):
...
+
+        self.output({"output_stack": output_stack, "missing_tilespecs_zs": missing_ts_zs})
+
+    def validate_apply_mipmaps(self, input_stack, output_stack, zvalues):
+        missing_ts_zs = []
+        
+        for z in zvalues:
+            input_rs = renderapi.resolvedtiles.get_resolved_tiles_from_z(input_stack, z, render=self.render)
+            output_rs = renderapi.resolvedtiles.get_resolved_tiles_from_z(output_stack, z, render=self.render)
+            if len(input_rs.tilespecs) != len(output_rs.tilespecs): # pragma: no cover
```
djkapner: why not raise an exception here and have the job fail? what happens here is that there are missing tiles and it is just logged to the output json.

mahaling: the json output was for the workflow to catch it and set the job state appropriately. raising a validation error after the first missing tileset would not allow for us to check the entire list of Zs. This behavior is not desired.

mahaling:
closing this without merging owing to the generalized stack transition checks PR.

PR #197, opened 2019-04-02
Rough align qc

mahaling:
Rough alignment QC.
Plots being generated are
1. IOU plot
2. Distortion plot


integration_tests/test_rough_align_qc.py
```diff
@@ -0,0 +1,92 @@
...
+
+    outjson = os.path.join(outdir, "output.json")
+    
+    mod = RoughAlignmentQC(input_data=ex, args=['--output_json', outjson])
+    mod.run()
+
+    # check if there is a pdf file
+    items = os.listdir(ex['output_dir'])
+
+    assert(any([n.endswith("pdf") for n in items]))
```
djkapner: is it possible to pull the output file names from mod.<output file name> and check that that specific file exists.
checking for any pdf or any html in the directory seems like it could be fooled.

rendermodules/em_montage_qc/schemas.py
```diff
@@ -79,3 +79,43 @@ class DetectMontageDefectsParametersOutput(argschema.schemas.DefaultSchema):
...
+        required=False,
+        default=10,
+        missing=10,
+        description="Pool size")
+    output_dir = Str(
+        required=False,
+        default=None,
+        missing=None,
+        description="temp filename to save fig")
+    out_file_format = Int(
```
djkapner: would a string argument be more clear here?

mahaling:
> you did not want to include the rough slicer in this?

You already have rough slicer in render-modules... We can include that, but that would not be in this PR... Coz that's going to add little more complications in terms of the output files and manually viewing them..

integration_tests/test_rough_align_qc.py
mahaling: added this check.

rendermodules/em_montage_qc/schemas.py
mahaling: It doesn't matter... First time users will always have to check the schema help descriptions for accepted values of the input. But changed this for anyway for our convenience.

djkapner:
> > you did not want to include the rough slicer in this?
> 
> You already have rough slicer in render-modules... We can include that, but that would not be in this PR... Coz that's going to add little more complications in terms of the output files and manually viewing them..

? oh, ok. I thought we had a whole meeting about this where you said you were including it in QC. I haven't looked at that code on months.

mahaling:
> > > you did not want to include the rough slicer in this?
> > 
> > 
> > You already have rough slicer in render-modules... We can include that, but that would not be in this PR... Coz that's going to add little more complications in terms of the output files and manually viewing them..
> 
> ? oh, ok. I thought we had a whole meeting about this where you said you were including it in QC. I haven't looked at that code on months.

Yes we did, but we discussed the output to be in ndviz as a stack with some markers to identify each row -z mapping. Will do that in the next PR...

rendermodules/em_montage_qc/rough_align_qc.py
```diff
@@ -0,0 +1,455 @@
...
+from shapely.ops import cascaded_union
+import multiprocessing as mp 
+import json
+import requests
+from functools import partial
+
+import tempfile
+import seaborn as sns
+
+from descartes.patch import PolygonPatch
```
RussTorres: I'm not sure where the descartes dependency comes from, but it seems to install with render-modules.  Still, it is good to list it explicitly in the requirements file.

rendermodules/utilities/pillow_utils.py
```diff
@@ -1,3 +1,13 @@
 from PIL import Image
 
 Image.MAX_IMAGE_PIXELS = None
+
+# imports needed for rough_align_qc
+import os
+import matplotlib
```
RussTorres: Can this be in another module, e.g. `matplotlib_utils` so that things importing pillow's image don't also have to import matplotlib.

rendermodules/em_montage_qc/schemas.py
```diff
@@ -79,3 +80,44 @@ class DetectMontageDefectsParametersOutput(argschema.schemas.DefaultSchema):
...
+        description="min z")
+    maxZ = Int(
+        required=True,
+        description="max z")
+    pool_size = Int(
+        required=False,
+        default=10,
+        missing=10,
+        description="Pool size")
+    output_dir = Str(
```
RussTorres: Is there a reason not to make this an `OutputDir`?

integration_tests/test_rough_align_qc.py
```diff
@@ -0,0 +1,97 @@
...
+
+@pytest.fixture(scope='module')
+def downsample_stack(render):
+    pre_rough_stack = "pre_rough_stack"
+    
+    renderapi.stack.create_stack(pre_rough_stack, render=render)
+    renderapi.client.import_tilespecs(pre_rough_stack, MONTAGE_SCAPES_TSPECS, render=render)
+    renderapi.stack.set_stack_state(pre_rough_stack, "COMPLETE", render=render)
+    yield pre_rough_stack
+    #render.run(renderapi.stack.delete_stack, pre_rough_stack)
```
RussTorres: These stacks should be cleaned up after the scope ends.

rendermodules/utilities/pillow_utils.py
```diff
@@ -1,3 +1,13 @@
...
+if os.environ.get('DISPLAY', '') == '':
+    matplotlib.use('Agg')
+
+import matplotlib.pyplot as plt, mpld3
```
RussTorres: I'm not too familiar with it, but it's not clear to me what mpld3 is doing as a dependency, as I don't see where else it is called.
In any case, PEP8 would suggest to use multiple lines for this.
```suggestion
import matplotlib.pyplot as plt
import mpld3
```

rendermodules/em_montage_qc/rough_align_qc.py
mahaling: added

rendermodules/utilities/pillow_utils.py
mahaling: created a new file

integration_tests/test_rough_align_qc.py
mahaling: uncommented the line

rendermodules/utilities/pillow_utils.py
mahaling: accepted.

rendermodules/em_montage_qc/schemas.py
RussTorres: @mahaling This was unclear -- I actually meant  using `argschema.fields.OutputDir` as the field type -- I think "output_dir" as the name is fine.

rendermodules/utilities/pillow_utils.py
RussTorres: This was later removed, so unresolving.  It looks like the build was failing because seaborn is imported before matplotlib and so the backend was not actually changed by `rendermodules.utilities.matplotlib_utils`
I think the only reason tests pass now is that the tests import from `rendermodules.utilities.pillow_utils` before seaborn is called and thus incidentally set the matplotlib backend.  Note, however, that this will not work in production as in e.g. a docker container the import statement shows:
```
# python -c "import rendermodules.em_montage_qc.rough_align_qc"
rendermodules/utilities/pillow_utils.py:9: UserWarning: 
This call to matplotlib.use() has no effect because the backend has already
been chosen; matplotlib.use() must be called *before* pylab, matplotlib.pyplot,
or matplotlib.backends is imported for the first time.

The backend was *originally* set to 'TkAgg' by the following code:
  File "<string>", line 1, in <module>
  File "rendermodules/em_montage_qc/rough_align_qc.py", line 13, in <module>
    import seaborn as sns
  File "/opt/conda/lib/python2.7/site-packages/seaborn/__init__.py", line 6, in <module>
    from .rcmod import *
  File "/opt/conda/lib/python2.7/site-packages/seaborn/rcmod.py", line 5, in <module>
    from . import palettes, _orig_rc_params
  File "/opt/conda/lib/python2.7/site-packages/seaborn/palettes.py", line 12, in <module>
    from .utils import desaturate, set_hls_values, get_color_cycle
  File "/opt/conda/lib/python2.7/site-packages/seaborn/utils.py", line 11, in <module>
    import matplotlib.pyplot as plt
  File "/opt/conda/lib/python2.7/site-packages/matplotlib/pyplot.py", line 71, in <module>
    from matplotlib.backends import pylab_setup
  File "/opt/conda/lib/python2.7/site-packages/matplotlib/backends/__init__.py", line 17, in <module>
    line for line in traceback.format_stack()


  mpl.use('Agg')

```

mahaling: changed the order... I guess the tests are passing now.

rendermodules/em_montage_qc/schemas.py
mahaling: that is a classic confusion .... Will change it again...

rendermodules/utilities/pillow_utils.py
```diff
@@ -1,3 +1,14 @@
 from PIL import Image
 
 Image.MAX_IMAGE_PIXELS = None
+
+# imports needed for rough_align_qc
```
RussTorres: These matplotlib imports should not be necessary now.

mahaling: done.

PR #198, opened 2019-04-10
Stack transition validation

mahaling:
validation of stacks during transition. This mainly checks for any missing tiles especially during apply mipmaps and apply lens correction. 
A generalized validation has been added to perform this check wherever necessary.


rendermodules/module/render_module.py
```diff
@@ -109,6 +109,19 @@ def output_tilespecs_to_stack(self, tilespecs, output_stack=None,
             output_stack, tilespecs, sharedTransforms=sharedTransforms,
             pool_size=pool_size, close_stack=close_stack, render=render)
 
+    def validate_tilespecs(self, input_stack, output_stack, z, render=None):
```
djkapner: I suggest calling this as optional from this function:
```
def output_tilespecs_to_stack(...validate=False...):
    if validate:
        self.validate_tilespecs(self.input_stack, self.output_stack)
```

rendermodules/module/render_module.py
```diff
@@ -109,6 +109,19 @@ def output_tilespecs_to_stack(self, tilespecs, output_stack=None,
...
+        missing_ts = []
+
+        render = self.render if render is None else render
+
+        input_rs = renderapi.resolvedtiles.get_resolved_tiles_from_z(input_stack, z, render=render)
+        output_rs = renderapi.resolvedtiles.get_resolved_tiles_from_z(output_stack, z, render=render)
+        if len(input_rs.tilespecs) != len(output_rs.tilespecs): # pragma: no cover
```
djkapner: maybe instead of checking just len() check tileids in a set overlap way

djkapner: this is what I meant here:
```
in_ids = set([t.tileId for t in input_rs.tilespecs])                                                                               
out_ids = set([t.tileId for t in output_rs.tilespecs])                                                                             
if in_ids != out_ids:
    etc.
```

rendermodules/module/render_module.py
djkapner: we punted on this after discussion

rendermodules/module/render_module.py
mahaling: ok... have added the changes

RussTorres:
What is the point of the "recent changes" commit?  I don't see any code differences, but it looks like a lot of files are entirely overwritten by it.  Why is this, and can we scrub it for the sake of commit history?

mahaling:
> What is the point of the "recent changes" commit? I don't see any code differences, but it looks like a lot of files are entirely overwritten by it. Why is this, and can we scrub it for the sake of commit history?

could be an effect of merge. I do not recall the changes made in that commit. If it helps, please scrub it.

PR #199, opened 2019-04-24
Deprecate matlab

mahaling:
Deprecating matlab functions from modules and tests... 

Suggestions are welcome if they can be better organized or have other ideas.


mahaling:
> So, where's the part that lets you turn off testing for `deprecated`?

That would have to be setup in setup.py which I do not want to do now before everyone completely signs off matlab totally. That change will not be in this PR. Doesn't make sense to have tests written for deprecated functions and NOT run them here.

PR #200, opened 2019-05-02
Faster mesh tests

djkapner:


djkapner:
could merge here or after into develop

integration_tests/test_mesh_lens_correction.py
```diff
@@ -60,15 +171,24 @@ def test_coarse_mesh_failure(render, tmpdir_factory):
...
 
     meshmod = MeshLensCorrection(
             input_data=example_for_input,
             args=['--output_json', outjson])
 
+    # will raise Exception because there aren't many matches
     with pytest.raises(MeshLensCorrectionException):
         meshmod.run()
 
+    rmtree(outdir)
```
RussTorres: Not sure why this is necessary when outdir is created by a `tempdir_factory` fixture

integration_tests/test_mesh_lens_correction.py
```diff
@@ -147,7 +267,9 @@ def test_make_mask(tmpdir_factory):
                 basename=None)
 
 
-def test_mesh_lens_correction(render, tmpdir_factory):
+def test_mesh_lens_correction(
+        render, tmpdir_factory, raw_lens_stack_2, raw_lens_matches_2):
+
     outdir = str(tmpdir_factory.mktemp("mesh_lens"))
     out_html_dir = outdir
     example_for_input = dict(example)
```
mahaling: I don't know how it was done previously in this test, but we do a deep copy in all our other tests. It would be confusion free on the input data if a deep copy is done here (and in subsequent tests).

djkapner: good point, these were quite old

integration_tests/test_mesh_lens_correction.py
djkapner: in my local test environment, these directories seem to persist

djkapner: if we're going to abandon python 2.x we could start using context manager:
```
with TemporaryDirectory() as output_dir:
   blah
```

RussTorres: Right, I see that now.  Would it make more sense to do a function-scoped fixture in this case, then?

djkapner: sure, i'll do that. it will be cleaner

PR #203, opened 2019-05-22
Deprecate trakem2 lens correction

RussTorres:
This introduces a module-level structure to the deprecated branch, moving the deprecated `montage` and `rough_align`  modules.  This also moves the trakem2 derivation component of `lens_correction` to deprecated  so we can later remove it from testing and code coverage.


rendermodules/deprecated/solver/schemas.py
```diff
@@ -1,10 +1,8 @@
 from argschema.fields import Bool, Float, Int, Nested, Str, InputDir, OutputDir
 from argschema.schemas import DefaultSchema
-import marshmallow as mm
-from marshmallow import Schema, validates_schema, post_load
-from ..module.render_module import RenderParameters
-from rendermodules.rough_align.schemas import *
-import os
+from rendermodules.rough_align.schemas import (
+    PointMatchParameters, PastixParameters,
```
djkapner: I'm not sure why PastixParameters would not be in a deprecated place.

djkapner: I guess that should have been part of the deprecate matlab PR

RussTorres: You're right, I was trying to minimize changes to the code while shuffling things around.  We may as well do it in this PR.

Issue #192, opened 2019-03-29
Deprecate MATLAB solver


RussTorres:
The EMaligner MATLAB solver has served us well but is a deployment/debugging nightmare and EM_aligner_python has now been battle tested by the largest EM dataset to date.

My suggestion is to keep the code to call the MATLAB solver (excluded from coverage) as well as tests related to them with a skip condition and duplicate those e.g. rough alignment test fixtures using the python solver.

RussTorres:
This will relieve some of the issues in #190

PR #204, opened 2019-05-23
Skip deprecated

### LINKED ISSUE #193: Deprecate TrakEM2 lens correction
The trakem2 lens correction script is largely not used in production and we do not do sufficient integration tests against e.g. FIJI versions, but it might make sense to keep the module readily available.  Similar to the MATLAB strategy in #192, I think we should keep the code on hand and avoid the slow tests against it.  This helps testing speed as discussed in #190.

RussTorres:
adds functionality to skip testing of legacy code (MATLAB solver, TrakEM2 lens correction) by default while allowing those tests to be run selectively with environment variables.  Also modifies the coveragerc to omit everything under deprecated.

This follows #203, and resolves #193 and #192

### DIFF
--- a/integration_tests/test_rough_align_matlab.py
+++ b/integration_tests/test_rough_align_matlab.py
@@ -19,7 +19,8 @@ from test_data import (
         test_rough_parameters as solver_example,
         apply_rough_alignment_example as ex1,
         rough_solver_example as solver_input,
-        pool_size)
+        pool_size,
+        test_legacy_rough)
 from rendermodules.module.render_module import RenderModuleException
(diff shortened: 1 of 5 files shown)


integration_tests/test_rough_align_matlab.py
```diff
@@ -30,6 +31,12 @@
 import shutil
 import numpy as np
 
+# skip these tests if not explicitly asked for
+pytestmark = pytest.mark.skipif(not test_legacy_rough, reason=(
```
djkapner: my first time seeing `pytest.mark.skipif`. This way for the module vs. decorating each function seems a little non-obvious, but... it seems to be how it's done

PR #206, opened 2019-06-11
Multi-channel support

perlman:
This is @fcollman's multichannel modifications from August 2018, with a few added tests.

The tests work locally, and we'll see soon if I copied the correct data to the test directory.

I am concerned about one function:
https://github.com/AllenInstitute/render-modules/blob/ep_multichan_tests/rendermodules/intensity_correction/apply_multiplicative_correction.py#L107

`corr_dict` is a new argument, thus breaking any existing use of this function. Is that something to worry about?  (If so, I can move it to the end as corr_dict={}.)

### DIFF
--- a/rendermodules/intensity_correction/apply_multiplicative_correction.py
+++ b/rendermodules/intensity_correction/apply_multiplicative_correction.py
@@ -168,8 +168,6 @@ class MultIntensityCorr(StackTransitionModule):
                 corr_dict[chan.name] = CC
         
         outdir = self.args['output_directory']
-        if not os.path.exists(outdir):
-            os.makedirs(outdir)
 
         mypartial = partial(
             process_tile,
--- a/rendermodules/intensity_correction/apply_multiplicative_correction.py
+++ b/rendermodules/intensity_correction/apply_multiplicative_correction.py
@@ -104,7 +104,7 @@ def write_image(dirout, orig_imageurl, Res, stackname, z):
     tifffile.imsave(outImage, Res)
     return outImage
 
-def process_tile_multichan(C, corr_dict, dirout, stackname, clip, scale_factor, clip_min, clip_max, input_ts):
+def process_tile(C, dirout, stackname, clip, scale_factor, clip_min, clip_max, input_ts, corr_dict=None):
     """function to correct each tile in the input_ts with the matrix C,
     and potentially move the original tiles to a new location.abs
 
@@ -144,11 +144,6 @@ def process_tile_multichan(C, corr_dict, dirout, stackname, clip, scale_factor, 
 
     return output_ts
 
-def process_tile(C, dirout, stackname, clip, scale_factor, clip_min, clip_max, input_ts):
-    """function to maintain the existing process_tile prototype"""
-    return process_tile_multichan(C, None, dirout, stackname, clip, scale_factor, clip_min, clip_max, input_ts)
-
-
 class MultIntensityCorr(StackTransitionModule):
     default_schema = MultIntensityCorrParams
 
@@ -177,15 +172,15 @@ class MultIntensityCorr(StackTransitionModule):
             os.makedirs(outdir)
 
         mypartial = partial(
-            process_tile_multichan,
+            process_tile,
             C,
-            corr_dict,
             self.args['output_directory'],
             self.args['output_stack'],
             self.args['clip'],
             self.args['scale_factor'],
             self.args['clip_min'],
-            self.args['clip_max'])
+            self.args['clip_max'],
+            corr_dict=corr_dict)
         with renderapi.client.WithPool(self.args['pool_size']) as pool:
             output_tilespecs = pool.map(mypartial, inp_tilespecs)
 
--- a/rendermodules/intensity_correction/apply_multiplicative_correction.py
+++ b/rendermodules/intensity_correction/apply_multiplicative_correction.py
@@ -7,22 +7,24 @@ import numpy as np
 import tifffile
 from ..module.render_module import RenderModule
 from ..module.render_module import StackTransitionModule
+from ..module.render_module import RenderModuleException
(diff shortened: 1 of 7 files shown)


perlman:
Ok, a minor change to preserve process_tile, but place all the code into process_tile_multichan, allowing backward compatibility.  (Setting a default value did not seem to be immediately compatible with the existing partial(process_tile,...) code.

rendermodules/intensity_correction/apply_multiplicative_correction.py
```diff
@@ -70,50 +72,82 @@ def getImage(ts):
...
         (presently assumes this is a tiff image whose URL can be read with tifffile)
-
+    channel: str
+        channel name to get image of, default=None which will default to the non channel image pyramid
     Returns
     =======
     numpy.array
         2d numpy array of this image
     """
-    d = ts.to_dict()
```
fcollman: was this d never used?

rendermodules/intensity_correction/apply_multiplicative_correction.py
```diff
@@ -70,50 +72,82 @@ def getImage(ts):
...
 
+def write_image(dirout, orig_imageurl, Res, stackname, z):
+    [head, tail] = os.path.split(orig_imageurl)
+    outImage = os.path.join("%s" % dirout, "%s_%04d_%s" %
+                            (stackname, z, tail))
+    tifffile.imsave(outImage, Res)
+    return outImage
 
-def process_tile(C, dirout, stackname, clip,scale_factor,clip_min,clip_max,input_ts):
+def process_tile_multichan(C, corr_dict, dirout, stackname, clip, scale_factor, clip_min, clip_max, input_ts):
```
fcollman: can you make this optional keyword argument?

rendermodules/intensity_correction/calculate_multiplicative_correction.py
```diff
@@ -109,11 +168,14 @@ def run(self):
...
+        # renderapi.stack.set_stack_state(
         #    self.args['output_stack'], "COMPLETE", render=self.render)
 
 
 if __name__ == "__main__":
     mod = MakeMedian(input_data=example_input)
     mod.run()
+
+
+mylist = [i*i for i in range(5) if i<4]
```
fcollman: I think this was an oversight

fcollman:
i have some minor comments, but given that the single channel tests are still passing I think this will work for both EM and multichannel use cases.

rendermodules/intensity_correction/apply_multiplicative_correction.py
perlman: It was an orphan when you added it in 905189524a11e81149f9737faf8702e1005f8d93.

rendermodules/intensity_correction/apply_multiplicative_correction.py
perlman: I assumed I couldn't, because of the partial, but forgot that partial can take a named argument. I'll adjust.

perlman: Making it optional is a bit messy with the call via partial().  It can't be the last argument -- that's reserved for the tilespec. I'm not sure what the easy solution is without reworking the execution.

rendermodules/intensity_correction/calculate_multiplicative_correction.py
perlman: Definitely

rendermodules/intensity_correction/apply_multiplicative_correction.py
fcollman: partial works fine with keyword arguments that come after... so you if you have f(a,b,c,d=None) and you do myp= partial(f, 1,2,d=5). then myp(4) is f(a=1, b=2, c=4,d=5)

perlman: I thought I had tried this already, but must have forgotten to change the order of arguments passed in. Swapped back to just process_tile(...corr_dict=None), and it seems to work.

rendermodules/intensity_correction/apply_multiplicative_correction.py
```diff
@@ -127,9 +156,31 @@ def run(self):
...
+
+        # construct a dictionary with the correction factors
+        corr_dict = {}
+        if corr_ts.channels is not None:
+            for chan in corr_ts.channels:
+                Nc, Mc, CC = getImage(corr_ts, chan.name)
+                corr_dict[chan.name] = CC
+        
+        outdir = self.args['output_directory']
+        if not os.path.exists(outdir):
```
RussTorres: making output_directory an `argschema.fields.OutputDir` in the schema will create this while also checking for directory mode and avoiding race conditions.

rendermodules/intensity_correction/calculate_multiplicative_correction.py
```diff
@@ -58,48 +96,69 @@ def run(self):
...
+        if channel_names[0] is not None:
+            try:
+                default_chan = next(ch.name for ch in firstts[0].channels if ch.ip[0].imageUrl == firstts[0].ip[0].imageUrl )
+            except StopIteration:
+                raise RenderModuleException("default channel doesn't match any channel")
+
         outdir = self.args['output_directory']
         if not os.path.exists(outdir):
             os.makedirs(outdir)
+        print("channel_name",channel_names)
```
RussTorres: This should be able to go to the argschemaparser logger

perlman: This doesn't need to exist, logger or not. Removing.

rendermodules/intensity_correction/apply_multiplicative_correction.py
perlman: Thanks for the pointer.

PR #207, opened 2019-06-11
Add close_stack to stack.consolidate_transforms

perlman:
This was a minor change @TotteKarlsson and I made a while ago to have a close_stack option when consolidating transformations.  (If this belongs elsewhere, please let me know.)


fcollman:
EM doesn't want to close stacks ever (always streaming processing) but I think its a useful option to have.

PR #209, opened 2019-07-11
External lens correction

djkapner:
* offloads lens correction solving to `em_stitch` repository, sharing the same algorithms with the scopes.
* jumps through some hoops to make sure the `opencv-python` and `opencv-contrib-python` requirements of the scopes,  the workflow, and `em_stitch` don't clobber each other. My solution here was to modify render-modules Dockerfile and this: https://github.com/AllenInstitute/em_stitch/blob/41871cd32c3fb206cec81067aae21778a686b8a9/setup.py#L48-L50
I am happy to modify either of them to make this work well.


integration_tests/test_mesh_lens_correction.py
```diff
@@ -155,6 +147,8 @@ def test_coarse_mesh_failure(
     example_for_input['rerun_pointmatch'] = False
     outjson = os.path.join(output_directory, 'mesh_lens_out0.json')
 
+    print(json.dumps(example_for_input, indent=2))
```
mahaling: can we change this to just json.dumps rather than a print statement. unless you require something to be printed

rendermodules/mesh_lens_correction/do_mesh_lens_correction.py
```diff
@@ -254,27 +233,49 @@ def run(self):
...
+                'compress_output': False,
+                'log_level': self.args['log_level']}
+
         meshclass = MeshAndSolveTransform(
-                input_data=args_for_input,
-                args=['--output_json', out_file.name])
+                input_data=solver_args, args=[])
         # find the lens correction, write out to new stack
         meshclass.run()
+        with open(meshclass.args['output_json'], 'r') as f:
```
RussTorres: using an `output_json`  populated within the MeshAndSolveTransform module based on the `output_dir` ( I think?) will make this module less robust to name collisions in cases where the same lens correction is double-run (which used to happen fairly often in automated processing, but may not at this point.)

rendermodules/mesh_lens_correction/do_mesh_lens_correction.py
```diff
@@ -254,27 +233,49 @@ def run(self):
...
+        with open(tform_out, 'w') as f:
+            json.dump(resolved.transforms[0].to_dict(), f)
 
         try:
             self.output(
-                    {'output_json': meshclass.args['outfile'],
-                     'maskUrl': self.maskUrl,
-                     'qc_json': out_file.name})
+                    {'output_json': tform_out,
+                     'maskUrl': self.maskUrl})
```
RussTorres: Are we abandoning lens correction QC altogether?   Was there any logic (computer or human) that previously depended on this?

djkapner: We had turned it off at some point because something in QC was hanging (some sort of tree was getting too large with the number of point matches in lens generation.) We got along fine without it. Initially, I had that because I thought this was effectively a montage, it should pass montage QC, but, we have other checks in place that seem better suited to catching problems with lens correction. So, yes, I propose we abandon it altogether.
I don't think anything depended on this being in the output json. I certainly never looked at it.

rendermodules/mesh_lens_correction/do_mesh_lens_correction.py
djkapner: I guess there are 2 things here:
the output_json could be overwritten, and, the resolved json.gz could be overwritten. If these were timestamped filenames, would that alleviate the concern?

rendermodules/mesh_lens_correction/do_mesh_lens_correction.py
RussTorres: Cool, let's drop it.

rendermodules/mesh_lens_correction/do_mesh_lens_correction.py
RussTorres: I'd be happy with that or being able to specify individual output files independent of the `output_dir`.

djkapner: v0.2.3 of `em_stitch` now has a timestamp boolean argument. I'm leaving the default as False, so as not to rock the microscope boat right now, but, hard-coding it to True here.
https://github.com/AllenInstitute/em_stitch/commit/c3439cf2a22703cd3c5c2daff620d53c05a334f8

integration_tests/test_mesh_lens_correction.py
djkapner: @mahaling can you please finish your review?

PR #210, opened 2019-07-12
making qc html file names informative

djkapner:
I think this will help communicating qc results on slack, and just general readability.
currently, qc output html files are name like:
`<out_html_dir>/tmpyc_ynqzh.html`
especially when debugging manually, one can get a directory full of these and it would be nice to easily pick out what section from the name.
This PR replaces with:
`<out_html_dir>/<stack>_<z>_<timestamp>.html`
`stack` and `z` are already known to the function that writes the files. `timestamp` is a millisecond-precision datetime.now() string just in case multiple nodes are writing output files for the same stack/z to the same directory. (generally probably not)

### DIFF
--- a/rendermodules/em_montage_qc/plots.py
+++ b/rendermodules/em_montage_qc/plots.py
@@ -2,6 +2,8 @@ import numpy as np
 import tempfile
 import renderapi
 import requests
+import os
+import datetime
 from functools import partial
 
 from rendermodules.residuals import compute_residuals as cr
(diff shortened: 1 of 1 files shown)


rendermodules/em_montage_qc/plots.py
```diff
@@ -136,10 +138,14 @@ def plot_defects(render, stack, out_html_dir, args):
...
-    out_html.close()
+    out_html = os.path.join(
+            out_html_dir,
+            "%s_%d_%s.html" % (
+                stack,
+                z,
+                datetime.datetime.now().strftime('%Y%m%d%H%S%M%f')))
 
-    output_file(out_html.name)
+    output_file(out_html)
```
mahaling: I am just wondering if this works... the integration tests doesn't check if the returned value is a html file name and not a tempfile object.

rendermodules/em_montage_qc/plots.py
```diff
@@ -220,7 +226,7 @@ def plot_defects(render, stack, out_html_dir, args):
 
     save(plot_tabs)
 
-    return out_html.name
+    return out_html
```
mahaling: same here... this has to return a filename not a file object.

rendermodules/em_montage_qc/plots.py
djkapner: it isn't a file object, it is the path:
https://github.com/AllenInstitute/render-modules/blob/c860bf7b4719c576b7804a7a6337d76f28052963/rendermodules/em_montage_qc/plots.py#L141-L146

rendermodules/em_montage_qc/plots.py
djkapner: same comment as above

PR #213, opened 2019-07-16
Label lens tform

djkapner:
when we pointmatch for montaging, our tilespecs have a transform list that looks like:
```
[ReferenceTransform(5c1529edb83474000163e84b),
 M=[[1.000000,0.000000],[0.000000,1.000000]] B=[216535.000000,231299.000000]]
```
and neither of our transforms are labelled. Somewhere deep in render, it just lops off the last transform and pointmatches in lens-corrected units. For example, here:
https://github.com/saalfeldlab/render/blob/907850047cc4ad22493754464d538cd31920ba52/render-app/src/main/java/org/janelia/alignment/spec/ListTransformSpec.java#L173-L222

For rough alignment, we typically have just one transform, and matching happens in raw units, which we want.

For fine alignment, or a more general case, we could have more than 2 transforms. Then, the matching gets screwed up. The simple fix is to label the lens correction transform as "lens" and then matching happens again in lens corrected space. This PR labels the lens correction transform as "lens" at the time of ApplyLensCorrection.

### DIFF
--- a/integration_tests/test_lens_correction.py
+++ b/integration_tests/test_lens_correction.py
@@ -97,6 +97,16 @@ def stack_lc(render):
     renderapi.stack.set_stack_state(stack, 'COMPLETE', render=render)
     yield stack
 
+
+@pytest.fixture(scope='module')
+def stack_lc_label(render):
+    stack = "test_lc_label"
+    renderapi.stack.create_stack(stack, render=render)
+    tspecs = [renderapi.tilespec.TileSpec(json=tspec)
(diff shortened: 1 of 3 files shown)


RussTorres:
For reference, before merging in: permalink to where this behavior is implemented https://github.com/saalfeldlab/render/blob/c3b4e95d02f642916eba9edceacf60857442aa35/render-ws/src/main/java/org/janelia/render/service/TileDataService.java#L310

PR #218, opened 2019-09-03
Uri inputs uh

RussTorres:
I finally got tests going, cut a release of the `uri_handler` and was able to update this branch to use that to seamlessly handle file and s3-like schemas.  this PR uses uris instead of posix paths for ingest, apply mipmaps, generate_tilespecs, generate_mipmaps, and apply_lens_correction.

since using uris is the new default behavior, all existing schemas for these modules should convert older inputs to uris, leaving tests untouched.


RussTorres:
Needs another commit to get rid of some commented out code and a once-over to make sure additional changes since the much-earlier commits don't reference file paths.

rendermodules/utilities/uri_utils.py
```diff
@@ -0,0 +1,89 @@
...
+     uri_prefix)
+
+from uri_handler.uri_functions import (
+    uri_readbytes,
+    uri_writebytes)
+
+__all__ = [
+    "uri_join", "uri_prefix", "uri_readbytes", "uri_writebytes"]
+
+# def join_paths(paths, delimiter='/'):
```
djkapner: are you leaving all these commented functions here intentionally?

RussTorres: I was not -- deleted now.

rendermodules/dataimport/create_mipmaps.py
```diff
@@ -94,10 +111,10 @@ def mipmap_PIL(im, levels_file_map, ds_filter="NEAREST",
 }
 
 
-def create_mipmaps(inputImage, outputDirectory='.', method="block_reduce",
-                   mipmaplevels=[1, 2, 3], outputformat='tif',
-                   convertTo8bit=True, force_redo=True,
-                   **kwargs):
+def create_mipmaps_uri(inputImage, outputDirectory=None, method="block_reduce",
```
mahaling: I hope outputDirectory defaults to some dir in the schema.

PR #223, opened 2019-10-06
Bug fix v2

mahaling:


rendermodules/rough_align/schemas.py
```diff
@@ -342,35 +342,32 @@ class DownsampleMaskHandlerSchema(RenderParameters):
         required=True,
         description="stack that is read and modified with masks")
     close_stack = argschema.fields.Bool(
-        require=True,
+        required=True,
```
RussTorres: This should not have a default if required (and other versions of this are not required)

Issue #251, opened 2023-12-01
use em-stitch version of generate_EM_tilespecs_from_metafile


RussTorres:
em-stitch is now a dependency of this repo and has a diverged version of `generate_EM_tilespecs_from_metafile`.  We should either subclass this to generate our tilespecs.

PR #267, opened 2025-01-13
non-render lens correction derivation module

RussTorres:
Adds a module for more integrated lens correction derivation.  So integrated that it does not involve communication with the render server, hence some digression from the usual module structure.

This code as it exists is very useful in some current workflows, so it will be preferentially merged -- I will open an issue for some known concerns.


RussTorres:
Also of note: one piece of code here is useful for the common interface for tilespec generation between em-stitch and asap-modules mentioned in #251

